PR #32, opened 2026-03-27
ops: Implemented Dockerfile and updated the compose.yaml to

Moritz2k1:
Implemented Dockerfile, updated compose.yaml to include backend service for deployment

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -31,4 +31,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: ./gradlew build sonar --info
+        run: ./gradlew build jacocoTestReport sonar --info
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ build/
 !**/src/main/**/build/
 !**/src/test/**/build/
 
-
 ### STS ###
 .apt_generated
 .classpath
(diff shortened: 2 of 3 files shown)


valentinaschiavon99 (changes requested):
Thank you for implementing Docker containerization: this is a solid foundation with good multi-stage build practices and proper service orchestration. A few items need attention before merging:

**What's Working Well** :
-Multi-stage build and proper base images
-Clean port exposure and compose integration
-Good environmental variable approach

**Issues to Address** :

- Security (High): Container runs as root. Add unprivileged user before ENTRYPOINT.
- Port mapping (High): ${SERVER_PORT:-8080}:8080 with SERVER_PORT in environment creates a mismatch. Either remove SERVER_PORT from environment or map both sides consistently.
- Environment defaults (Medium): DB_NAME, DB_USERNAME, DB_PASSWORD need :- defaults to prevent empty strings.
- Gradle build (Medium): Use ./gradlew bootJar -x test instead of build -x test to avoid JaCoCo failures.
- JAR reference (Medium): Change *-SNAPSHOT.jar to *.jar to support future releases.
- .dockerignore (Low): Add to exclude unnecessary files from build context.
- Database readiness (Medium): Add health check to postgres with condition: service_healthy in depends_on.

Please review the Copilot suggestions in the comments: they provide implementation details for these items. Test locally with docker compose up once updated. Great work on this foundational piece! 👍🏻

eles17 (approved):
All previously requested changes have been addressed 👍 

#### Non-blocking note: 
- later we could also add a backend healthcheck and a short README section for docker usage.

Anyhow, the setup now looks consistent and ready for use. Good job addressing the feedback and improving the configuration.

PR #39, opened 2026-04-03
Feat/database schema

### LINKED ISSUE #36: Implement Database Schema
### **Goal**

Implement Machi Koro database schema from UML diagram and verify the tables in pgAdmin4.
The goal is not to finish all of the tables and entities, but to show others how the process works.
Since, we do not have a full-fledged perfect UML diagram, only the first few basic tables should be implemented.

### **Tasks**

- Review UML diagram and identify entities and relationships
- Setup Jetbrains Exposed ORM
- Configure database connection
- Define table objects
- Create and query tables

### **Definition of Done**

- [x] Exposed ORM is functioning
- [x] Connection to docker container is working
- [x] First table objects are defined
- [x] Querying of the tables works

Moritz2k1:
Tables are implemented under the database folder. This should give people a general guideline how to implement tables in the database. I cannot guarantee if the tables are correct themselves, but it is a first start and fine tuning can be done later. Next step is to setup swagger + endpoints to do CRUD methods


Moritz2k1:
Updated the code, deleted the dotenv tests, since they shouldn't be relevant anymore (at least i hope so).
Added database connection tests

FabianSolutions (dismissed):

valentinaschiavon99 (changes requested):
Overall: Solid first step,Moritz! The table definitions are clean and give the team a clear guideline to follow. A few issues need to be addressed before this is ready to merge.

Critical — .env support is broken for local development The dotenv bootstrapping was removed from main(), but the app still throws IllegalArgumentException at startup if DB_USERNAME or DB_PASSWORD are not available as real OS environment variables. In practice this means the app only starts when Docker Compose injects them, and fails silently for anyone running it locally without a .env file. The previous ignoreIfMissing() approach handled this cleanly and should be restored.

connectToDatabase() should not be a top-level function Hard to mock and test in isolation. Move it into a dedicated @Configuration class.

Unused imports in ServerApplication.kt Dotenv, count, insert, select, selectAll are never referenced. Clean them up.

println instead of a logger Use SLF4J for startup messages, consistent with the rest of the codebase.

Incomplete test cleanup in ServerApplicationTests restoreProperties() should call clearProperty when the original value was null, otherwise test state leaks between runs.
After these changes it can be merged :)

eles17 (changes requested):
I agree with review from @valentinaschiavon99 . The main issue is not the schema itself, but that the PR currently makes local development and review harder.

Also, non-database tests should not depend on Docker/Testcontainers. Right now this makes simple review harder for teammates who just want to run regular tests locally. 
The README/local setup should also clearly explain what requires Docker and what should still work without it.

Moritz2k1:
Just merge this please. I am so done.

valentinaschiavon99 (approved):
**APPROVED**
Ran full test suite - all tests pass with coverage requirements met.
The PR adds a complete database schema for the Machi Koro game with proper table definitions, domain models, and initialization logic. Code is clean, well-tested, and follows project conventions.
Schema design is solid with appropriate foreign key relationships, cascade deletes, and sensible defaults. The Testcontainers setup ensures reliable database testing in CI. Docker Compose configuration is ready for local development.
Tested manually with Lev using Docker - database initializes correctly and works as expected.
No blocking issues found. Ready to merge.
Nice work Moritz!

eles17 (approved):
Valentina and me tested both with the Docker-based setup and a local PostgreSQL setup, and both worked. Docker was more consistent and much easier to get started with, so I’d recommend that everyone install Docker on their laptop. It makes the project easier to run, simplifies debugging, and avoids having to set up the database manually from scratch.

The branch looks good from my side, so I’m approving the PR.

PR #94, opened 2026-04-16
feat: add GamePhaseService for turn phase state machine

FabianSolutions:
Sub-issue of #46, closes #84

Adds a stateless `GamePhaseService` that implements the Machi Koro turn phase cycle.

#### Changes

- New `GamePhaseService` in `service/` package with two methods:
  - `nextPhase(currentPhase)` — returns the next phase in the cycle
  - `initialPhase()` — returns `ROLL_DICE` as the starting phase
- Phase cycle: `ROLL_DICE` → `RESOLVE_EFFECTS` → `BUY_OR_BUILD` → `END_TURN` → `ROLL_DICE`

#### Verified

- 7 unit tests covering all individual transitions, exhaustiveness, and full cycle
- All existing tests still pass

PR #105, opened 2026-04-19
Feat/85 persist turn phase

FabianSolutions:
Adds the persistence layer for turn phases and wires it into the phase state machine merged in #94.


Moritz2k1:
Is it possible to drop the repository/ folder and move the methods of the GameRepository under the GameDao class?

Moritz2k1 (approved):
This is how I imagined it to be. The code looks very clean and as long as the tests pass, I fully believe in the correctness

PR #123, opened 2026-04-24
83 generate unique lobby code

Ivona-Goranova:
Added LobbyCodeGenerator for generating 7-character uppercase alphanumeric lobby codes.

Changes:
- Added LobbyCodeGenerator in domain/utils
- Added test for generated code format


FabianSolutions (changes requested):
Blockers                                                                                                                                                                                                                          
                                                                            
  - Dead code. GameDao.create() still uses the inline (1000000..9999999).random().toString() — the new generator is never called. Needs lobbyCode = LobbyCodeGenerator.generate().
  - Not actually "unique". generate() is only random. No DB unique index on Games.lobbyCode, no collision retry. Title/AC unmet.

  Should fix

  - Format change (digits → alphanumeric) is a client-visible contract — coordinate before wiring.
  - Single-shot random test — loop ~100 iterations; tighten predicate to it in 'A'..'Z' || it in '0'..'9'.

  Nits

  - Missing trailing newline on LobbyCodeGenerator.kt.
  - Import order in DomainTest.kt (utils placed before models).
  - Random.Default is probably fine; SecureRandom if griefing-by-guess is in scope.

Request changes — utility is fine, but the ticket isn't done until the DAO is switched over and uniqueness is actually enforced.

Moritz2k1 (approved):
The GameDao has been extended by implementing a lobby code when creating a new game that is random. Tests also have been changed accordingly

PR #134, opened 2026-04-25
Feat/133 earnings controller

### LINKED ISSUE #133: Implement EarningsController
### Goal
Implementation of the EarningsController that will be used to communicate with the frontend.

Moritz2k1:
Implemented the EarningsController that will be used to communicate between the backend and frontend for the earnings logic

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/EarningsController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/EarningsController.kt
@@ -11,7 +11,7 @@ import org.springframework.messaging.simp.SimpMessagingTemplate
 import org.springframework.stereotype.Controller
 
 /**
- * Handles earnings resolution triggered by the frontend after a dice roll
+ * Handles earnings resolution triggered by the frontend after a dice roll.
  */
 @Controller
 class EarningsController(
@@ -21,15 +21,16 @@ class EarningsController(
     private val logger = LoggerFactory.getLogger(EarningsController::class.java)
 
     /**
-     * Resolves card effects for all players and notifies everyone in the game
+     * Resolves card effects for all players and broadcasts the result to the game topic.
      */
     @MessageMapping("/game.resolveEffects")
     fun resolveEffects(@Payload request: ResolveEffectsRequest) {
+        val gameTopic = "/topic/game/${request.gameId}"
         try {
             earningsService.resolveEffects(request.gameId)
             logger.info("Resolved effects for game ${request.gameId}")
             messagingTemplate.convertAndSend(
-                "/topic/public",
+                gameTopic,
                 WebSocketMessage(
                     type = MessageType.GAME_ACTION,
                     sender = "server",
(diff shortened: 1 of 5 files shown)


valentinaschiavon99 (dismissed):
Thank you for the contribution. This pull request introduces the EarningsController, which is responsible for triggering the earnings resolution phase of the game. The code is clean, follows existing patterns, and includes valuable unit tests.
Here is a summary of what works well and what should be addressed before this PR is merged.
What Works Well
•
Clear Responsibility: The controller correctly delegates the core logic to the EarningsService, keeping the controller lean and focused on handling requests.
•
Asynchronous Communication: The use of SimpMessagingTemplate to send WebSocket messages is appropriate for notifying clients of game state changes in real-time.
•
Test Coverage: The EarningsControllerTest provides good coverage for both the success case and a scenario where the service layer throws an exception. The use of mockito-kotlin is effective.
What Needs to Be Changed
1.
**Critical Bug: Broadcasting to a Public Topic**
◦
Issue: The controller sends the EFFECTS_RESOLVED message to /topic/public. This is a critical bug. It means that all players in all active games will receive this message, regardless of which game it originated from. This will lead to incorrect client-side behavior and data leaks between games.
◦
Required Change: The destination topic must be made specific to the game instance. It should be changed from /topic/public to a dynamic topic like /topic/game/${request.gameId}.
2.
**Poor User Experience: Silent Error Handling**
◦
Issue: The try-catch block currently catches any Exception, logs it, and then stops. The client who initiated the request receives no feedback that the operation failed. From the user's perspective, they will click a button and nothing will happen.
◦
Required Change: When an exception occurs, a WebSocket message should be sent back to the requesting user (or to a game-specific error topic) to inform them of the failure. This ensures the user interface can react appropriately.
**Conclusion**
While the foundation of this feature is solid, the broadcasting bug is too critical to ignore. The PR cannot be approved in its current state.

Moritz2k1:
I think this issue:
"Issue: The controller sends the EFFECTS_RESOLVED message to /topic/public. This is a critical bug. It means that all players in all active games will receive this message, regardless of which game it originated from. This will lead to incorrect client-side behavior and data leaks between games." is also present in the other controllers. Please double check this in the other files as well

valentinaschiavon99:
> I think this issue:
> 
> "Issue: The controller sends the EFFECTS_RESOLVED message to /topic/public. This is a critical bug. It means that all players in all active games will receive this message, regardless of which game it originated from. This will lead to incorrect client-side behavior and data leaks between games." is also present in the other controllers. Please double check this in the other files as well
> 
> 
> 
> 

That's a good idea! Will do that.
We can create an issue about it

Moritz2k1:
I have implemented the requested changes

FabianSolutions (approved):
LGTM. Clean controller, good split between service/messaging, and the per-game topic (/topic/game/{gameId}) is the right call instead of broadcasting to all of /topic/public. Tests cover both the happy and the error path with a proper argument captor — nice.

Issue #61, opened 2026-04-13
enforce purchase rules and add edge case tests


eles17:
## Goal

Harden purchase-related game rules and verify correct behavior through focused edge case testing.

## Tasks

- [x] Ensure player cannot perform more than one purchase per turn
- [x] Ensure invalid purchase attempts do not modify game state
- [x] Verify purchased establishments are correctly available for future activation
- [x] Add edge case tests (e.g. insufficient coins, invalid card, repeated purchase attempts)
- [x] Refactor or adjust existing logic if inconsistencies are found

## Definition of Done

- [ ] Second purchase in the same turn is rejected reliably
- [ ] Invalid purchase attempts do not mutate coins, player state, or supply
- [ ] Purchased cards behave correctly in subsequent turns
- [ ] Edge case scenarios are covered with tests
- [ ] All tests pass and no regressions are introduced
- [ ] No crashes or invalid state transitions occur
- [ ] Works in Docker environment
- [ ] PR created and CI passes

### Related issues / waiting on / depends on

Depends on: Backend - purchase handling  
Parent issue: #45

PR #139, opened 2026-04-26
60 implement purchase handling validation and execution

### LINKED ISSUE #60: Implement purchase handling (validation and execution)
## Goal

Implement complete backend handling for purchasing an establishment or landmark, including validation and execution.

## Tasks

- [ ] Validate player has enough coins for the selected purchase
- [ ] Validate selected card/landmark is available in supply
- [ ] Enforce rule: only one purchase per turn
- [ ] Deduct coins from player on successful purchase
- [ ] Add purchased card/landmark to player state
- [ ] Remove purchased card from supply
- [ ] Return proper error response for invalid purchase attempts
- [ ] Add tests for valid and invalid purchase scenarios

## Definition of Done

- [ ] Valid purchase correctly updates coins, player state, and supply
- [ ] Invalid purchase (e.g. not enough coins, invalid card, second purchase) is rejected
- [ ] Game state remains consistent after invalid attempts
- [ ] No crashes or invalid state mutations occur
- [ ] Tests for purchase handling are implemented and passing
- [ ] Works with WebSocket (request/response flow)
- [ ] Works in Docker environment
- [ ] PR created and CI passes

### Related issues / waiting on / depends on

Depends on: Backend - buying phase flow  
Parent issue: #45

eles17:
#### Summary
Adds the first working backend purchase flow for BUY_OR_BUILD.

#### What  Changed

- Added /app/game.purchase
- Added purchase handling for one establishment or one landmark per turn
- Deducts coins and updates player/supply state on success
- Tracks whether the player has already purchased this turn
- Resets purchase state on turn change

#### Validation

- Rejects purchases outside BUY_OR_BUILD
- Rejects second purchases in the same turn
- Rejects insufficient-coins, unavailable-card, and already-built-landmark cases

Further hardening in #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,20 @@ Exposed entities are object-oriented wrappers around a single database row. They
 via delegated properties and handle the persistence mechanics internally. Entities are strictly internal to the database
 layer — they are always converted into domain models before being passed to services or controllers.
 
+### Domain Models
+
+Domain models represent the data used in the application's core business logic.
+
+They are pure Kotlin data classes that encapsulate business state independently of persistence frameworks and transport layers. These models are used primarily by the service layer to implement game rules and manage the overall game state.
+
+Key characteristics:
+
+- Encapsulate pure business data independent of persistence and frameworks
+- Used by the service layer to implement game logic
+- Represent the current game state and rule-related state transitions
+- Do not contain database logic or persistence concerns
+- Can be used as a foundation for DTOs when appropriate, though DTOs should remain transport-focused
+
 ### Data Transfer Objects (DTOs)
 
 DTOs are simple objects used exclusively to carry data between the server and clients. They define the shape of requests
--- a/README.md
+++ b/README.md
@@ -1,404 +1,183 @@
-# Server
+# Machi Koro Server
 
-## Getting Started
+> Robust real-time multiplayer backend for the Machi Koro board game, built with Kotlin, Spring Boot, and WebSockets.
 
-### Setup Environment Variables
+This server manages real-time Machi Koro game sessions, player communication, game state, and persistent storage on
+PostgreSQL. Designed for reliability, scalability, and developer productivity.
 
-1. Rename `.env.example` to `.env`
-2. Configure database credentials:
-   ```env
-   DB_HOST=localhost
-   DB_PORT=5432
-   DB_NAME=machi_koro
-   DB_USERNAME=your_username
-   DB_PASSWORD=your_password
-   SERVER_PORT=8080
+## Key Features
+
+- **Real-Time Multiplayer:** Instant bidirectional communication using STOMP over WebSockets.
+- **Game State Management:** Strict tracking of turn phases (Roll Dice, Resolve Effects, Buy/Build, End Turn) and game
+  statuses.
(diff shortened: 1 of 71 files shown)


Moritz2k1:
Please resolve the conflicts, then we can merge this

Moritz2k1:
> Please resolve the conflicts, then we can merge this

I merged the main branch into this feature branch and resolved the conflicts as good as I could - all tests pass, which is a good sign

FabianSolutions (changes requested):
Should fix before merge
                                                                                                                                                                                                                                                                                                                    
  1. Purchase isn't atomic. purchaseEstablishment runs updateCoins, upsert, decrementQuantity, and updateHasPurchasedThisTurn as four separate transactions. If one fails partway through, you can end up with deducted coins but no card, or a card without the supply being decremented. purchaseLandmark has the
  same shape. Wrap the whole flow in one transaction { … } (or @Transactional on the service method) so it commits or rolls back as a unit.
  2. Race on hasPurchasedThisTurn. Two messages arriving in quick succession (double-click, client retry) can both read false, both pass the guard, and both go through. The check-then-write needs to happen inside the same transaction with a row lock, or be expressed as a conditional update.
  3. No caller authorization. The handler trusts gameId from the payload and assumes the sender is the active player — any connected client could trigger a purchase on someone else's behalf. The PR notes hardening lives in #61, which is fine, but please leave a // TODO(#61) at the entry point so it doesn't
  silently slip.

  Minor

  4. purchaseEstablishment(... cardType: CardType?, landmarkType: LandmarkType?) takes both nullable params and then rejects the wrong shape inside. Cleaner to validate the shape once in purchase() and pass only the non-null value down.
  5. gameMarketplaceDao.initForGame(gameId) being called lazily in the purchase path hides a dependency and adds a write to every purchase. If the marketplace should exist for every in-progress game, initialize it at game start.
  6. Missing test for INVALID_PURCHASE_REQUEST — e.g. ESTABLISHMENT with both fields set, or neither. The branch exists; worth covering.
  7. A handful of files end without a trailing newline (GameController.kt, GameModel.kt, Tables.kt, PurchaseServiceTest.kt, …). Style nit.
  8. broadcastPurchase uses linkedMapOf<String, String> while the other broadcasts use mapOf. Ordering isn't meaningful here — mapOf would match the surrounding code.

  Nice

  - PurchaseResult is a clean return shape.
  - hasPurchasedThisTurn is correctly reset in advanceTurn, with a test that proves it.
  - Good unit/integration test coverage for the happy paths and most rejection cases.

Approve once #1 and #2 are handled.

FabianSolutions (approved):
Good to merge.

eles17:
Hey @Moritz2k1 & @FabianSolutions , quick note regarding this PR.

This was my assigned issue/PR and I was just about to  resolve the conflicts.

I understand that resolving conflicts to unblock a merge can be helpful, especially if something is time-sensitive. In that case, **please just check with the owner of the PR first!**

In this situation, there was no urgency and I was about to finish it myself. Making changes and merging without any consultation breaks ownership and can lead to confusion.

Going forward, let’s align on this:
If a PR is assigned to someone, please ask before modifying or resolving its conflicts yourself - unless it is explicitly urgent.

Thanks 👍

PR #144, opened 2026-04-28
feat: return individual dice values in RollDiceResponse

kutschera13:
- Add RollDiceResponse DTO with dice list and total fields
- Update DiceService to roll and return individual dice values instead of sum
- Update WebSocketController to send dice array in payload
- Fix DiceServiceTests to use getPlayers and assert RollDiceResponse
- Fix WebSocketControllerTests to include DiceService mock
- Fix ChatMessageTests to include new MessageType values (PLAYER_LEFT_FINISHED_GAME, ERROR, ROLL_DICE)


Moritz2k1 (commented):
Please do not remove comments (WebSocketController.kt)
Otherwise it's fine - I will add the comments back again, so we can merge this

Moritz2k1 (approved):
THe RollDiceRequest and RollDiceResponse were implemented, GameController has the rollDice method now, DiceService looks good.  Bugs fixed in the GameController - methods now send in the game.id instead of lobby-wide

FabianSolutions (approved):
Lookgs good, merging.

PR #147, opened 2026-04-28
Ref/145 simplify db

### LINKED ISSUE #145: Simplify Database Layer
### Goal

Right now the Database Layer uses the Exposed ORM with both approaches (DSL and Entity API). Since, this is overly complicated I will look into the best practices for a SpringBoot project and come up with a unified, simple solution

### Tasks

- [ ] Look into the best practices for a SpringBoot project
- [ ] Identify the files that need to be changed
- [ ] Switch fully to the Entity API or the DSL approach OR come up with something different
- [ ] Update ReadME
- [ ] Adjust tests

Moritz2k1:
This PR is meant to simplify the database schema/db layer.
The entities have been removed and instead of mixing the Entity API and DSL approach - I have switched it fully to DSL now.
The ReadMe also got updated.
I know that this PR is quite big, but all tests passed without adjusting the code.


FabianSolutions (changes requested):
Overview                                                                                                                                                                                                                          
   
  Drops Exposed's Entity API in favor of DSL-only access. Deletes 5 entity classes, rewrites the 5 DAOs to use selectAll/where/insertAndGetId/update/deleteWhere, with a private ResultRow.toModel() in each DAO. README updated.   
  All tests still pass unchanged.                                 

  Strengths

  - Consistent pattern across DAOs; no Exposed types leak out.
  - updateTurnPhase / updateHasPurchasedThisTurn now actually throw GameNotFoundException when no row is updated — small correctness win.
  - README rewrite is clearer than the old "DSL vs Entity" section.

  Issues

  1. Lost domain error messages on FK violations. Old GameDao.create and PlayerDao.addPlayer did UserEntity.findById(...) ?: error("User not found"). New code inserts the FK directly, so a bad hostUserId/gameId/userId now
  surfaces as a raw ExposedSQLException instead of a clean domain error. Either restore the existence check or translate the FK violation in the service layer.

  2. PlayerDao.addPlayer is multi-transaction. countByGameId → insertAndGetId → findById are three separate transactions. Concurrent joins can produce duplicate turnOrder. Pre-existing, but worth wrapping in a single transaction
   { } (and ideally a unique constraint on (gameId, turnOrder)).

  3. Inconsistent missing-row handling. Two updates throw; the rest (updateStatus, updateAfterRoll, advanceTurn, updateCoins, updateTurnOrder, updateSessionToken, delete) silently no-op. Pick one policy.

  4. README nits. The cp .env.example .env fence lost its list indent so it'll render outside step 2; the table separator on line 131 has stray leading whitespace.

  5. Missing trailing newlines in GameDao.kt, PlayerDao.kt, UserDao.kt.

  Tests

  Public contract preserved, but the new throw-on-missing branches and the changed error-message behavior aren't covered. Worth 1–2 tests to lock them in.


Solid refactor. Address #1 before merging; #2 is a good cleanup while you're here. Rest are nits.

FabianSolutions (approved):
The twi most important issues are properly addressed. Ready to merg.

FabianSolutions (dismissed):
Re-reviewed at `b117d27`.

The two most important issues from my first pass are addressed:

- **Atomicity** — `PlayerDao.addPlayer` is now wrapped in a single `transaction { }`; count, insert, and final select all run together. The final lookup also avoids the previous `findById(playerId)!!` round-trip.
- **Consistent missing-row handling** — every update/delete in `GameDao`, `PlayerDao`, `UserDao` now checks the affected-row count and throws a domain-specific exception (`GameNotFoundException`, new `PlayerNotFoundException`, new `UserNotFoundException`). The two new exception types are placed under `org.machikoro.server.exception/` next to the existing one — consistent.

Net improvement, no regressions. Approving.

### Follow-ups (not blocking)

- Add a unique index on `Players(gameId, turnOrder)` to fully close the TOCTOU race on `turnOrder` — the single transaction makes the race atomic, but at default Postgres `READ COMMITTED` isolation two concurrent inserts can still both see count=N. A unique constraint forces the second insert to fail.
- Add tests covering the new throw branches (15+ new throw sites; not exercised by the existing suite).
- Restore explicit FK-existence checks in `GameDao.create(hostUserId)` and `PlayerDao.addPlayer(gameId, userId)` if you want clean domain errors for bad references — currently a bad FK still surfaces as a raw `ExposedSQLException`.

### Cosmetic

- Trailing newlines missing on `GameDao.kt`, `PlayerDao.kt`, `UserDao.kt`.
- README: the bash fence inside the numbered list is still un-indented (will likely render outside step 1); the table separator on line 131 has stray leading whitespace.

Worth a confirmation that the latest CI run on `b117d27` is green before merging.

FabianSolutions (approved):

PR #153, opened 2026-04-29
146 refactor broadcast game end event

### LINKED ISSUE #146: refactor: Broadcast Game End Event
User Story
As the server, I want to notify all players when the game ends so that clients can display the winner screen.

Acceptance Criteria

- [ ] Event sent to all players
- [ ] Winner info included
- [ ] Event sent once

doniliks:
Overview
This PR improves how a turn ends and how the game detects a winner.
As assumed by current code, endTurn is called after purchase that allows player to change card to purchase. After call of endTurn, current turn phase is set to end_turn. If there are winner no future actions can be done except sending winner back to the client, otherwise values are updated to handle next player's turn.

What changed

- advancePhase can no longer be called from Client to ensure correct flow. Now it should be called after any successful phase
- advanceTurn became a private function and is called only in endTurn to ensure that next player starts only after current player is finished with his turn

- Better turn handling:
  endTurn now returns a result (EndTurnOutcome) instead of just a phase. It clearly shows if the game continues, or if someone won
- Counters of played games or/and wins are incremented after finishing game
- Winner check moved to service. 
   Winner detection is now inside GamePhaseService. Removed this logic from the controller
- Cleanup after game ends: 
   Marketplace is cleared
   Player cards are deleted
   Player landmarks are deleted
Players and game itself are not deleted at this point since they can be reused to rejoin current lobby.
- Winner can only be checked in the correct phase (END_TURN)
- new DAO methods: delete all cards for a player
delete all landmarks for a player
- EndTurnOutcome:
Continue if game continues
Won if game finished with a winner


FabianSolutions (changes requested):
Nice direction overall — collapsing winner detection into endTurn and returning a sealed EndTurnOutcome is a real improvement over the controller-driven flow. There is one blocking bug and a handful of smaller items to address
   before this can merge.                                         

  **Must fix**

  1. Wrong ID passed to incrementWins / incrementGamesPlayed

  In GamePhaseService.kt:

  winConditionService.detectWinner(gameId)?.let { winner ->
      userDao.incrementWins(winner.id)   // winner.id is PlayerModel.id, not the user ID
      finishGame(gameId)
      return EndTurnOutcome.Won(winner)
  }

  …and in finishGame:

  playerDao.getPlayers(gameId).forEach { userDao.incrementGamesPlayed(it.id) }  // same problem

  UserDao.incrementWins / incrementGamesPlayed look up Users.id, but PlayerModel.id is the player row ID, not the user ID. In production this will either throw UserNotFoundException or silently increment the stats of an
  unrelated user. Replace with winner.userId and it.userId.

  The unit tests don't catch this because userDao is mocked. Please add a test (ideally integration-level, or at minimum one where PlayerModel.id != userId) so this can't regress.

  **Should fix before merging**

  2. Stale Swagger documentation
  OpenApiConfig.kt still advertises /app/game.advancePhase, but you removed that mapping. Drop that row from the endpoint table.

  3. Redundant ensureGameIsRunning call
  endTurn already calls gameStateGuard.ensureGameIsRunning(gameId), then detectWinner calls it again. It doubles the DB hit and couples WinConditionService to the guard for what was a pure read. Consider passing the
  already-loaded GameModel (or just the phase) into detectWinner.

  4. clearDBAfterGame is not atomic
  gameMarketplaceDao.deleteAllForGame, playerCardDao.deleteAllByPlayerId, and playerLandmarkDao.deleteAllByPlayerId each open their own transaction. If one fails midway, the game is left half-cleaned with status already set to
  FINISHED. Wrap the cleanup (or the full finishGame path) in a single transaction { ... }.

  5. Leftover TODO in production code
  // TODO() Prove if needed since advanceTurn always sets ROLL_DICE as current phase in GamePhaseService.kt — either resolve it or remove the comment. // TODO() also looks like the Kotlin TODO() function, which is easy to
  misread.

  **Minor**

  6. KDoc copy-paste
  PlayerLandmarkDao.deleteAllByPlayerId says "Deletes all card entry for a player" — should say "landmarks". The PlayerCardDao version should also be "cards" (plural).

  7. Misleading test name
  `advancePhase wraps END_TURN back to ROLL_DICE` no longer calls advancePhase; it calls endTurn. Rename it so failures are easier to interpret.

  8. Formatting
  - gameDao.updateTurnPhase(gameId,TurnPhase.END_TURN) — missing space after the comma.
  - private fun broadcastPhase(gameId: Int, newPhase : TurnPhase) — stray space before the colon.
  - PlayerCardDao.kt, PlayerLandmarkDao.kt, and GamePhaseServiceTest.kt are missing trailing newlines.

  9. Naming
  clearDBAfterGame reads awkwardly and leaks "DB" into a service-layer method. Something like cleanupFinishedGameData would fit better.

  10. Wildcard import
  GamePhaseServiceTest.kt uses import org.mockito.kotlin.* while the rest of the test suite imports explicitly. Worth keeping consistent.

doniliks:
Hi Fabian, thanks for your feedback.

I noticed that the winner check has been moved to immediately after purchase without broadcasting landmark purchase at all, directly moving to winner broadcast. With this change, my previous updates with EndTurnOutcome no longer have practical use, and it’s partly meaningless to make changes based on your feedback since parts of the related logic have been modified or removed, e.g . "Consider passing the already-loaded GameModel (or just the phase) into detectWinner." since you're passing gameId only.

I think we should first align on whether the endTurn function should remain, if we want to allow cases where a player can reconsider purchase and change it before confirming it with end turn.

Also, from what I can see, the purchase is still being broadcast, but the turn itself is not currently advancing, since endTurn was also responsible for this if there was no winner in this phase.

In this case we could keep end turn function but call it internally after buying phase. 

Let me know how you’d like to proceed it.

valentinaschiavon99:
let us know when this PR is ready for review!

doniliks:
should be ready now

FabianSolutions (approved):
Mentioned issues addressed

Moritz2k1 (approved):
Approved, since all the changes have been implemented, request by Fabian

PR #154, opened 2026-04-30
feat: #130 add REST companion controller for WebSocket actions

### LINKED ISSUE #130: Add a REST companion controller for WebSocket actions
Create a @RestController marked with @Hidden or @Operation that exposes the same actions as POST endpoints purely for Swagger testing. These call the same service methods as the WebSocket controllers.

FabianSolutions:
Closes #130.

## Summary
- New `GameRestController` at `/api/dev/game/**` mirrors the `/app/game.*` STOMP actions (`purchase`, `endTurn`, `leave`, `rollDice`, `resolveEffects`) as plain HTTP POST endpoints, delegating to the same service methods so behaviour stays identical. Intended for Swagger UI / curl / Postman driven testing.
- Gated behind `@Profile("!prod")` and grouped under a dedicated `WebSocket Companion (dev)` Swagger tag. `SecurityConfig` adds `/api/dev/**` to `permitAll` and to the CSRF ignore list so the endpoints are actually reachable from curl/Postman without a session-bound token — please review that tradeoff.
- `/app/game.advancePhase` is intentionally **not** mirrored: PR #153 removes that STOMP mapping, so adding a companion for it now would just need to be ripped out. The stale row in the `OpenApiConfig` WebSocket table is also dropped here.

## Coordination with #153
When #153 merges first, this branch will need a small rebase: `gamePhaseService.endTurn` will return `EndTurnOutcome` instead of `TurnPhase`, and winner detection moves inside `endTurn`. The companion's `endTurn` handler (and its tests) will need to be reshaped to consume the sealed result.

## Test plan
- [x] `GameRestControllerTest` — 11 new unit tests cover the success and failure path of every endpoint, plus both branches of `endTurn` (winner / no winner). All pass.
- [x] Full `./gradlew test` green except for 18 pre-existing testcontainer-dependent failures that are also failing on `main` in this environment (no Docker locally).
- [ ] Reviewer to spot-check `/swagger-ui.html` shows the new `WebSocket Companion (dev)` tag.
- [ ] Reviewer or local QA to exercise at least one endpoint end-to-end against a running game (e.g. `POST /api/dev/game/rollDice`).
- [ ] Confirm endpoints are absent when `SPRING_PROFILES_ACTIVE=prod`.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/config/OpenApiConfig.kt
+++ b/src/main/kotlin/org/machikoro/server/config/OpenApiConfig.kt
@@ -21,18 +21,21 @@ class OpenApiConfig {
                 .description(
                     """
     REST and WebSocket API for the Machi Koro board game.
-    
+
     ## WebSocket Connection
     Connect via STOMP at ws://localhost:8080/ws.
-    
+
     ### Endpoints
     | Destination | Payload | Broadcasts to |
     |---|---|---|
     | /app/game.resolveEffects | ResolveEffectsRequest | /topic/public |
-    | /app/game.advancePhase | AdvancePhaseRequest | /topic/public |
     | /app/game.endTurn | EndTurnRequest | /topic/public |
     | /app/chat.send | WebSocketMessage | /topic/public |
     | /app/chat.addUser | WebSocketMessage | /topic/public |
+
+    For Swagger-driven testing of these actions over plain HTTP, see the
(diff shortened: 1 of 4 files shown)


Moritz2k1 (commented):
We do not need a REST GameController + Tests. These things should just be noted down in the documentation (OpenAPIConfig). I think the issue is ultimately obsolete, since we voted for just documenting the REST stuff

Issue #162, opened 2026-04-30
Add registration REST endpoint


FabianSolutions:
### Goal

Expose a registration endpoint so clients can create a new user with a username and password. Required for the username+password auth flow chosen for #157.

### Tasks

- [x] `POST /auth/register` accepting `{ username, password }`.
- [x] Reject usernames that already exist (`UserDao.findByUsername(...)` returns non-null) with a clear error.
- [x] Hash the password via the `PasswordEncoder` introduced in #161 and persist via `UserDao.create(...)`.
- [x] Allow the new endpoint in `SecurityConfig.permitAll(...)`.
- [x] Decide a minimal validation contract — at least non-empty username and non-empty password. Anything stricter (length, charset, complexity) can be a follow-up.
- [x] Unit / integration tests for the happy path, duplicate-username path, and validation failures.

### Out of scope

- Login / token issuance — see #158.
- Password strength rules, email verification, account recovery — file follow-ups if desired.
- Any client-side UI work — that goes in the Client repo as a separate issue once this endpoint is stable.

### Acceptance

- A `POST /auth/register` with a fresh username + password succeeds and persists the user with a hashed password.
- A second registration with the same username returns a 4xx error and does not mutate state.
- An empty username or empty password returns a 4xx error.

### Related

Parent: #157
Depends on: #161

Issue #158, opened 2026-04-30
Add login REST endpoint that issues a session token


FabianSolutions:
### Goal

Expose a login REST endpoint that authenticates a user via **username + password** and issues a session token. Currently `Users.sessionToken` and `UserDao.updateSessionToken` / `findBySessionToken` exist, but no controller writes or returns a token, so clients have no way to obtain one.

Credential model: **username + password** (decided 2026-04-30). Password storage and registration are tracked separately — see #161 and #162.

### Tasks

- [ ] `POST /auth/login` accepting `{ username, password }`; on success generates a fresh session token, persists it via `UserDao.updateSessionToken`, and returns it.
- [ ] Verify the password against the stored hash using the `PasswordEncoder` introduced in #161.
- [ ] Reject unknown usernames or wrong passwords with the **same** generic error response — do not let a caller distinguish "user does not exist" from "wrong password".
- [ ] `POST /auth/logout` that clears the current token (`UserDao.updateSessionToken(id, null)`).
- [ ] Allow both endpoints in `SecurityConfig.permitAll(...)`.
- [ ] Unit / integration tests: happy-path login, wrong-password, unknown-username, logout invalidates token.

### Out of scope

- Wiring the token into the WebSocket handshake — see #159.
- Using the token to gate game endpoints — see #160.
- Token expiration / rotation — file as a follow-up if needed.

### Acceptance

- A registered client can `POST /auth/login` with valid credentials and receive a non-null `sessionToken`.
- A subsequent `POST /auth/logout` clears the token, after which `UserDao.findBySessionToken(token)` returns null.
- Login with an unknown username and login with a wrong password produce the same error response.

### Related

Parent: #157
Depends on: #161
Related: #162 (registration is required before any client can log in, but the two endpoint PRs can be developed in parallel once #161 is done)

PR #163, opened 2026-04-30
feat: #161 add password storage to Users (passwordHash + BCrypt bean)

### LINKED ISSUE #161: Add password storage to Users (passwordHash column + BCrypt hashing)
### Goal

Foundation for username+password authentication. Currently `Users` only stores `username`, `sessionToken`, and stats — there is no password column. Both the registration and login endpoints depend on this schema change being in place first.

### Tasks

- [x] Add `passwordHash varchar(60)` (or appropriate length for BCrypt) to `Users` in `Tables.kt`. Make it nullable initially to keep the migration safe; consider tightening to non-null in a follow-up once all users have one.
- [x] Wire a `PasswordEncoder` bean (Spring Security's `BCryptPasswordEncoder` — no extra Gradle dep needed since `spring-boot-starter-security` is already on the classpath).
- [x] Update `UserDao`:
  - `create(username)` → take a `passwordHash` parameter and persist it.
  - Add `updatePasswordHash(id, hash)` if needed.
  - `UserModel` gets a `passwordHash: String?` field, but `findBy*` methods should consider whether to expose it (probably keep it on the model but don't return it from any controller).
- [x] Tests: round-trip a user through `create` + `findById` and assert the hash is persisted; assert that two `BCryptPasswordEncoder.encode(same-password)` calls produce different hashes but both verify.

### Out of scope

- The login and registration endpoints themselves — separate issues.
- Password strength rules, rate-limiting, lockouts — file separately if we want them.
- Migrating existing users (none expected at this stage of the project).

### Acceptance

- A user can be created with a hashed password.
- `passwordEncoder.matches(rawPassword, user.passwordHash)` returns true for the password used at creation.
- All existing `UserDao` tests still pass.

### Related

Parent: #157

FabianSolutions:
Closes #161. Foundation for #158 (login) and #162 (registration).

## Summary
- Add a nullable `password_hash varchar(60)` column to `Users` (BCrypt's fixed output length) and surface it on `UserModel`. The field is annotated with `@field:JsonIgnore` so Jackson cannot accidentally serialize it into a response body, even if a future controller returns `UserModel` directly.
- Wire a `BCryptPasswordEncoder` bean in `SecurityConfig` (no Gradle change — `spring-boot-starter-security` is already on the classpath).
- Extend `UserDao`: `create(username, passwordHash = null)` with an optional default so existing test fixtures keep working unchanged, plus a new `updatePasswordHash(id, passwordHash)` mirroring the shape of `updateSessionToken`.
- Swap schema initialisation to `SchemaUtils.createMissingTablesAndColumns` so the new column appears on the next startup against existing databases. The function is deprecated in Exposed v1 in favour of a real migration tool — accepted for SE2 scope, suppressed with an inline comment pointing at a follow-up to introduce Flyway/Liquibase.

## Out of scope
- Login and registration endpoints — separate issues #158 and #162.
- Tightening `passwordHash` to non-null — follow-up after #162 lands and every real user has a hash.
- Password strength rules, lockouts, rate limiting — file separately if wanted.

## Test plan
- [x] `./gradlew clean compileKotlin compileTestKotlin` — clean, no warnings.
- [x] `SecurityConfigTests` passes locally — covers bean wiring, BCrypt salt randomisation, and the encode/verify round-trip including a wrong-password rejection.
- [x] Full `./gradlew test` shows the same 18 testcontainer-dependent failures as `main` — no new breakage. The 4 new `UserDaoTest` cases cannot run locally without Docker; they will execute in CI.
- [x] Reviewer to confirm `UserDaoTest` passes in CI (round-trip with hash, default-null path, `updatePasswordHash` happy path, `updatePasswordHash` `UserNotFoundException` path).
- [x] Reviewer to spot-check that `password_hash` shows up as `varchar(60)` nullable on the next deployment against an existing DB.

## Commits
1. `feat: #161 add password_hash column and read it through UserDao` — schema, model, `toModel()`, `create()` optional param, round-trip tests.
2. `feat: #161 add BCryptPasswordEncoder bean` — `SecurityConfig` bean + 3 tests.
3. `feat: #161 add updatePasswordHash mutator on UserDao` — new method + 2 tests.
4. `fix: #161 polish from self-review` — `@JsonIgnore` on the model field, `assertFalse` cleanup, parameter rename, deprecation comment.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/config/SecurityConfig.kt
+++ b/src/main/kotlin/org/machikoro/server/config/SecurityConfig.kt
@@ -4,6 +4,8 @@ import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.security.config.annotation.web.builders.HttpSecurity
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder
+import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.security.web.SecurityFilterChain
 
 @Configuration
@@ -42,4 +44,7 @@ class SecurityConfig {
 
         return http.build()
     }
+
+    @Bean
+    fun passwordEncoder(): PasswordEncoder = BCryptPasswordEncoder()
 }
--- a/src/main/kotlin/org/machikoro/server/dao/UserDao.kt
+++ b/src/main/kotlin/org/machikoro/server/dao/UserDao.kt
@@ -86,6 +86,18 @@ class UserDao {
     }
 
     /**
+     * Replaces the stored password hash for a user
+     * Caller is responsible for hashing the raw password via PasswordEncoder
+     * before passing it in
+     */
+    fun updatePasswordHash(id: Int, hash: String): Unit = transaction {
+        val updatedRows = Users.update({ Users.id eq id }) {
+            it[Users.passwordHash] = hash
+        }
+        if (updatedRows == 0) throw UserNotFoundException("User $id not found")
+    }
+
+    /**
      * Increments user's total win count by 1
      */
     fun incrementWins(id: Int): Unit = transaction {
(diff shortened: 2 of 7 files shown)


Moritz2k1 (approved):
The User table got extended by implementing the password hash property. BCrypt als got correctly implemented  in the SecurityConfig. UserModel adjusted. Tests implemented. Merged

PR #164, opened 2026-04-30
feat: #162 add POST /auth/register endpoint

### LINKED ISSUE #162: Add registration REST endpoint

FabianSolutions:
Closes #162. Builds on #161; unblocks #158 (login).

## Summary
- New `AuthController` with `POST /auth/register` taking `{username, password}` and returning `{id, username}` on success. Mounted at `/auth/register` (no `/api` prefix, consistent with `/lobby`).
- New `AuthService.register` runs `findByUsername` + `create` inside `@Transactional` so a duplicate inserted between the two calls rolls back rather than leaks through. Hashes the raw password via the `BCryptPasswordEncoder` bean introduced in #161.
- New `DuplicateUserException`, `RegisterRequest`, `RegisterResponse` DTOs. `RegisterResponse` deliberately omits `passwordHash` / `sessionToken` / stat counters so the response can't grow new fields by accident — same defensive reasoning that produced `@field:JsonIgnore` on `UserModel.passwordHash` in #161.
- `SecurityConfig` permits `/auth/**` and adds it to the CSRF ignore list. `/auth/**` (not `/auth/register`) is forward-compatible so #158 can add login/logout without touching this file.

## Validation contract
- Username is **trimmed and lower-cased** before validation, dedup, and persistence — `"Alice"`, `"alice"`, and `"  alice  "` all map to the same account. Stops two visually-identical accounts from co-existing because the DB unique index is byte-for-byte.
- Username must be non-blank and ≤ 50 characters (matches the column width).
- Password must be non-blank and ≤ 72 characters. The 72-cap is BCrypt's silent-truncation point — we reject up front for a clear error rather than letting a long password "work" with only the first 72 bytes mattering.
- Password is **not** trimmed; surrounding whitespace stays significant.

## Out of scope
- Login / logout / token issuance — #158.
- Password strength rules, lockouts, rate limiting, account recovery — file follow-ups if wanted.
- Client-side UI — Client repo issue once #162 and #158 are stable.
- Replacing the controller's `catch (e: Exception)` blanket with a `@ControllerAdvice` — same pattern as `LobbyController`; worth a separate cross-cutting issue.

## Test plan
- [x] `./gradlew clean compileKotlin compileTestKotlin` — clean, no warnings.
- [x] `AuthServiceTest` — 7 unit tests pass: happy path, duplicate, blank username, blank password, oversized username, trim+lowercase normalization, oversized password.
- [x] `AuthControllerTests` — 3 unit tests pass: success, duplicate response, validation-failure response.
- [x] `SecurityConfigTests` — still passes after the broadened CSRF ignore list and new permit rule.
- [x] Full `./gradlew test` — same 18 testcontainer-dependent failures as `main`, no new ones.
- [x] **Manual end-to-end:** registered users via Swagger UI against the `compose-dev.yaml` Postgres, verified rows in pgAdmin with BCrypt-shaped `password_hash` populated. Confirmed:
  - 200 + `{id, username}` on happy path.
  - 400 + `Username '<x>' is already taken` on duplicate.
  - 400 + `Username must not be blank` on empty username.
  - 400 + `Password must be at most 72 characters` on a 73-char password.
  - `"  Alice  "` registers as `alice`.
  - WARN log on failure shows `username + reason`, INFO log on success — no password ever logged.
- [ ] Reviewer to spot-check the new `Authentication` tag in Swagger UI.

## Heads-up for reviewers
This PR widens CSRF for `/auth/**`. That's a deliberate call so non-browser clients (Android, curl, Postman) can register without a session-bound CSRF token. Documented in the KDoc on `securityFilterChain`. If you'd rather keep CSRF strict, the alternative is requiring clients to fetch a token first — happy to revisit.

## Heads-up for whoever pulls this branch locally
`SchemaUtils.create` (chosen in #161) does not add missing columns to existing tables, so a long-lived dev volume with a pre-#161 `users` table needs either `docker compose down -v` or a one-shot `ALTER TABLE users ADD COLUMN password_hash VARCHAR(60)`. CI testcontainers always start fresh so they're unaffected.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/config/SecurityConfig.kt
+++ b/src/main/kotlin/org/machikoro/server/config/SecurityConfig.kt
@@ -17,14 +17,19 @@ class SecurityConfig {
      * Permits unauthenticated access to WebSocket endpoints, static resources, the root/index page,
      * and the `/actuator/health` endpoint so Docker, CI, and monitoring can poll it without credentials.
      * Requires authentication for API endpoints.
-     * CSRF is disabled for WebSocket endpoints to allow SockJS fallback HTTP POST requests.
+     * CSRF is disabled for WebSocket endpoints to allow SockJS fallback HTTP POST requests
+     * and for the public auth endpoints since clients (e.g. the Android app) post JSON
+     * without a session-bound CSRF token.
      * CSRF remains enabled for all other endpoints including API routes.
      */
     @Bean
     fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
         http
             .csrf { csrf ->
-                csrf.ignoringRequestMatchers("/ws", "/ws/**", "/ws-sockjs", "/ws-sockjs/**")
+                csrf.ignoringRequestMatchers(
+                    "/ws", "/ws/**", "/ws-sockjs", "/ws-sockjs/**",
+                    "/auth/**",
+                )
             }
             .authorizeHttpRequests { auth ->
                 auth.requestMatchers("/", "/index.html", "/css/**", "/js/**", "/webjars/**").permitAll()
@@ -38,6 +43,7 @@ class SecurityConfig {
                     "/v3/api-docs",
                     "/v3/api-docs/**"
                 ).permitAll()
+                auth.requestMatchers("/auth/**").permitAll()
                 auth.requestMatchers("/api/**").authenticated()
                 auth.anyRequest().authenticated()
             }
--- a/src/main/kotlin/org/machikoro/server/controller/AuthController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/AuthController.kt
@@ -1,0 +1,27 @@
+package org.machikoro.server.controller
+
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.tags.Tag
+import org.machikoro.server.dto.RegisterRequest
+import org.machikoro.server.service.AuthService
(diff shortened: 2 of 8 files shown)


Moritz2k1 (approved):
Clear implementation of a Rest Authentication Controller + Service

Issue #160, opened 2026-04-30
Enforce active-player guard on game-mutating WebSocket endpoints


FabianSolutions:
### Goal

Once a `Principal` is attached per WS session (#159), every game-mutating handler should verify that the sender is the active player for the given `gameId` before mutating state. Currently any authenticated user could send `/app/game.endTurn` or `/app/game.purchase` for any game and the server would accept it.

### Affected endpoints

- `/app/game.purchase`
- `/app/game.endTurn`
- `/app/game.advancePhase` *(if still present after #153 lands)*
- `/app/game.rollDice`
- `/app/game.resolveEffects`
- `/app/game.leave` — the player must match their own session

### Approach

- Add a guard helper, e.g. `GameStateGuard.ensureSenderIsActivePlayer(gameId, principal)`, that loads the game's active player and compares against `principal.userId`.
- Throw `CustomWebSocketException("NOT_YOUR_TURN", ...)` on mismatch — already routed to `/user/queue/errors` by `GlobalWebSocketExceptionHandler`.
- Call from each handler (or via a `ChannelInterceptor` that inspects the destination prefix and short-circuits before reaching the controller).

### Side-cleanup

`PurchaseService.kt:38` has a stale `TODO(#61): Verify the websocket caller is the active player for this game.` — issue #61 is actually about purchase rule edge cases, not auth. Update that TODO to point to this issue (or to #157) as part of the implementing PR.

### Acceptance

- User A sending `/app/game.endTurn` with a `gameId` where User B is the active player gets a `NOT_YOUR_TURN` error and no state mutation occurs.
- The active player can still take their turn.
- Tests cover the rejected and accepted paths per affected endpoint.

### Related

Parent: #157
Depends on: #159

Issue #159, opened 2026-04-30
Authenticate WebSocket sessions and attach Principal


FabianSolutions:
### Goal

Today any client that opens a `/ws` connection can send any STOMP frame to any `@MessageMapping` — there is no concept of "who is sending this". The session has no `Principal`, and `WebSocketController.addUser` just trusts whatever username is in the payload.

This issue adds the foundation: validate a session token on STOMP CONNECT, look up the user, and attach a `Principal` to the WebSocket session so downstream handlers can read it.

### Tasks

- [ ] Add a `ChannelInterceptor` (or `HandshakeInterceptor`) that reads the session token from the STOMP CONNECT headers.
- [ ] Validate the token via `UserDao.findBySessionToken(...)`; reject the CONNECT frame if missing or invalid.
- [ ] On success, attach a `Principal` (with at least `userId` and `username`) to the WS session.
- [ ] Ensure existing chat / game tests still pass — test setup will need to send a token on connect.

### Out of scope

- Issuing the token — see #158.
- Enforcing `principal.userId == activePlayer.userId` per game endpoint — see follow-up issue.
- Reworking the existing chat handlers to use `Principal` instead of payload-supplied usernames. Nice-to-have, but not blocking.

### Acceptance

- A WS connection without a valid token is rejected at CONNECT.
- A connection with a valid token has a non-null `Principal` accessible from any `@MessageMapping`.
- Tests cover both the rejected-handshake and accepted-handshake paths.

### Related

Parent: #157
Depends on: #158

Issue #88, opened 2026-04-13
Store lobbies in memory


Ivona-Goranova:
## Goal
Store created lobbies in memory for later access.

## Tasks
- [ ] Create in-memory storage (Map)
- [ ] Store lobby using its code as key

## Definition of Done
- [ ] Lobby is retrievable by code
- [ ] No crashes
- [ ] Tested manually
- [ ] PR created and CI passes

Ivona-Goranova:
This functionality is already covered by the current implementation.

Lobbies are not stored in-memory, but persistently in the database as `Game` entities with status `WAITING`.

- Each lobby is represented by a Game
- The lobbyCode uniquely identifies the lobby
- Players can join via the lobbyCode
- Data is persisted via DAO layer (GameDao, PlayerDao)

Using the database as the single source of truth is more robust than in-memory storage, especially for reconnect scenarios.

Issue #92, opened 2026-04-13
WebSocket endpoint and broadcast


Ivona-Goranova:
**Goal**
Allow joining a lobby via WebSocket and notify clients.

**Tasks**
- [x] Add @MessageMapping("/lobby.join")
- [x] Call joinLobby()
- [x] Validate authenticated user and lobby code payload
- [x] Broadcast LOBBY_JOINED message to clients
- [x] Add controller tests for success and error cases

PR #167, opened 2026-05-01
feat: #158 add POST /auth/login and /auth/logout endpoints

### LINKED ISSUE #158: Add login REST endpoint that issues a session token

FabianSolutions:
Closes #158. Builds on #161 (password storage) and #162 (registration); unblocks #159 (WebSocket auth + Principal) and the Client-side login UI that follows Client #88.

## Summary
- New `AuthService.login` and `AuthService.logout`, both `@Transactional`. Reuses the existing `UserDao` (`findByUsername`, `findBySessionToken`, `updateSessionToken`) and `BCryptPasswordEncoder` bean introduced in #161 — no new dependencies, no schema changes.
- New `POST /auth/login` and `POST /auth/logout` on `AuthController`.
- New `InvalidCredentialsException` plus `LoginRequest` / `LoginResponse` / `LogoutRequest` DTOs (Swagger-annotated).

## Validation contract
- Username is trimmed and lower-cased before lookup, matching `register`'s normalization. `"Alice"` / `"alice"` / `"  alice  "` all authenticate the same identity.
- Username and password must be non-blank — same `IllegalArgumentException` → 400 path as `register`.
- Logout requires a non-blank session token. Unknown tokens are an idempotent no-op (200 with empty body), not 4xx — preserves the no-enumeration property.

## No-enumeration contract
- Unknown username, wrong password, and user-with-null-`passwordHash` all collapse into a single throw site with the exact same `InvalidCredentialsException("Invalid username or password")`.
- `AuthServiceTest` asserts the **message is identical** between the unknown-user and wrong-password paths.
- `AuthControllerTests` adds a regression test that asserts the **statusCode AND body are identical** at the HTTP boundary, so a future refactor can't silently re-introduce enumeration.

## Logging hygiene
- Logged: usernames (canonical, post-normalization) at INFO on success and WARN on failure.
- **Never logged:** raw passwords, session tokens.
- Logout's success log lives in the service (`Cleared session for user '<username>'`) so it carries the actual username; the controller no longer double-logs on the success path.
- Stale-token logout produces no log line (silent no-op).

## Out of scope (per issue body)
- **Timing-attack mitigation** — when the user doesn't exist we skip the BCrypt comparison, so an attacker measuring response time could distinguish unknown-user from wrong-password. Mitigation would be to always run a BCrypt comparison against a dummy hash. File as a follow-up if the team cares before #160.
- **Token expiration / rotation** — current tokens never expire and aren't refreshed.
- **Per-IP / per-user rate limiting on `/login`** — separate hardening issue.
- **WebSocket handshake reading the token** — #159.
- **Active-player guard on game endpoints** — #160.
- **`@ControllerAdvice` for global error handling** — same broad-`catch` pattern as `register`; cross-cutting cleanup, separate issue.

## Test plan
- [x] `./gradlew clean compileKotlin compileTestKotlin --warning-mode all` — clean, no warnings, no deprecation notes.
- [x] `AuthServiceTest` — 17 tests pass (7 register + 10 new login/logout including the no-enumeration assertion).
- [x] `AuthControllerTests` — 8 tests pass (3 register + 5 new login/logout including the HTTP-level no-enumeration regression test).
- [x] `./gradlew test --rerun-tasks` — full suite green end-to-end (with Docker running locally for testcontainers).
- [x] **Manual end-to-end via Swagger** against the local Postgres:
  - Register `charlie` / `hunter2` → 200.
  - Login `charlie` / `hunter2` → 200 + `{ sessionToken: <uuid>, username: "charlie" }`. Verified the row in pgAdmin shows the persisted token.
  - Login `charlie` / wrong password → 401 + `Invalid username or password`.
  - Login `ghost_user_does_not_exist` / anything → 401 + same `Invalid username or password` (verified the body string is byte-for-byte identical to the wrong-password response).
  - Login `""` / anything → 400 + `Username must not be blank`.
  - Logout with the issued token → 200 with empty body. pgAdmin confirms `session_token` is now `NULL`.
  - Logout with the same now-stale token → 200 with empty body, silent — no log line, no DB write.
  - Logout with `""` → 400 + `Session token must not be blank`.
- [x] Server log review — usernames present, no passwords, no tokens.

## Heads-up for reviewers
- 401 vs 400 is a deliberate split: 401 = "your credentials are wrong" (generic message preserves no-enumeration); 400 = "your request is malformed" (specific message acceptable since it's about request shape, not credentials).
- Logout returning 200 for unknown tokens is intentional. If a reviewer would rather have 404 for unknown tokens, that re-introduces token enumeration via status code — happy to discuss.

## Commit split (3)
1. `feat: #158 add InvalidCredentialsException and Login/Logout DTOs` — pure scaffolding.
2. `feat: #158 add AuthService.login and logout with normalized lookup` — service + 10 unit tests + helper extraction.
3. `feat: #158 add POST /auth/login and /auth/logout controller endpoints` — controller + 5 controller tests including HTTP-level no-enumeration regression.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/AuthController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/AuthController.kt
@@ -2,9 +2,13 @@ package org.machikoro.server.controller
 
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.tags.Tag
+import org.machikoro.server.dto.LoginRequest
+import org.machikoro.server.dto.LogoutRequest
 import org.machikoro.server.dto.RegisterRequest
+import org.machikoro.server.exception.InvalidCredentialsException
 import org.machikoro.server.service.AuthService
 import org.slf4j.LoggerFactory
+import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
@@ -26,6 +30,34 @@ class AuthController(private val authService: AuthService) {
             ResponseEntity.ok(response)
         } catch (e: Exception) {
             logger.warn("Registration failed for '{}': {}", request.username, e.message)
+            ResponseEntity.badRequest().body(e.message)
+        }
+    }
+
+    @PostMapping("/login")
+    @Operation(summary = "Authenticate and receive a session token")
+    fun login(@RequestBody request: LoginRequest): ResponseEntity<Any> {
+        return try {
+            val response = authService.login(request.username, request.password)
+            logger.info("Logged in user '{}'", response.username)
+            ResponseEntity.ok(response)
+        } catch (e: InvalidCredentialsException) {
+            logger.warn("Login failed for '{}': invalid credentials", request.username)
+            ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(e.message)
+        } catch (e: Exception) {
+            logger.warn("Login failed for '{}': {}", request.username, e.message)
+            ResponseEntity.badRequest().body(e.message)
+        }
+    }
+
+    @PostMapping("/logout")
+    @Operation(summary = "Invalidate the given session token")
+    fun logout(@RequestBody request: LogoutRequest): ResponseEntity<Any> {
+        return try {
+            authService.logout(request.sessionToken)
+            ResponseEntity.ok().build()
+        } catch (e: Exception) {
+            logger.warn("Logout failed: {}", e.message)
             ResponseEntity.badRequest().body(e.message)
         }
     }
--- a/src/main/kotlin/org/machikoro/server/domain/models/UserModel.kt
+++ b/src/main/kotlin/org/machikoro/server/domain/models/UserModel.kt
@@ -7,6 +7,7 @@ data class UserModel(
     val username: String,
     @field:JsonIgnore
     val passwordHash: String?,
+    @field:JsonIgnore
     val sessionToken: String?,
     val totalWins: Int,
     val totalGamesPlayed: Int
--- a/src/main/kotlin/org/machikoro/server/dto/LoginRequest.kt
+++ b/src/main/kotlin/org/machikoro/server/dto/LoginRequest.kt
@@ -1,0 +1,11 @@
+package org.machikoro.server.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Request to authenticate an existing user — sent to POST /auth/login")
+data class LoginRequest(
+    @Schema(description = "Username — case-insensitive, leading/trailing whitespace ignored", example = "alice")
+    val username: String,
+    @Schema(description = "Raw password — server compares to the stored BCrypt hash", example = "hunter2")
+    val password: String,
+)
--- a/src/main/kotlin/org/machikoro/server/dto/LoginResponse.kt
+++ b/src/main/kotlin/org/machikoro/server/dto/LoginResponse.kt
@@ -1,0 +1,11 @@
+package org.machikoro.server.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Response from a successful login — caller stores the session token and uses it for subsequent authenticated calls")
+data class LoginResponse(
+    @Schema(description = "Opaque session token. Treat as a secret. Echo back to /auth/logout to invalidate.", example = "550e8400-e29b-41d4-a716-446655440000")
+    val sessionToken: String,
+    @Schema(description = "Canonical (trimmed and lower-cased) username of the authenticated user", example = "alice")
+    val username: String,
+)
--- a/src/main/kotlin/org/machikoro/server/dto/LogoutRequest.kt
+++ b/src/main/kotlin/org/machikoro/server/dto/LogoutRequest.kt
@@ -1,0 +1,9 @@
+package org.machikoro.server.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Request to invalidate a session token — sent to POST /auth/logout")
+data class LogoutRequest(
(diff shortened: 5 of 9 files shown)


Moritz2k1 (dismissed):
Implementation of the Login and Logout behaviour. Adjusts the AuthController, added Schemas for the dto's

valentinaschiavon99 (dismissed):
I’ve gone through the PR for the Auth module. We have a couple of architectural concerns—specifically a timing side-channel that’s leaking user existence—that need to be resolved before we can merge this to main.

Here is the breakdown of what needs to be addressed.

---

## 🔴 CRITICAL: Timing Side-Channel (Username Enumeration)
**File:** `src/main/kotlin/org/machikoro/server/service/AuthService.kt`

The current logic returns immediately if the user isn't found in the database, but proceeds to the `passwordEncoder` if the user exists. Since BCrypt is intentionally slow (~100ms), an attacker can enumerate valid usernames by measuring response times. We need to ensure the password encoder is invoked in every execution path using a dummy hash to normalize the latency.

---

## 🟠 MEDIUM: Potential Session Token Leak
**File:** `src/main/kotlin/org/machikoro/server/domain/models/UserModel.kt`

While `passwordHash` is correctly ignored, `sessionToken` is missing the `@JsonIgnore` annotation. If this model is ever serialized directly in a future endpoint, we risk leaking active session tokens. We need to ensure this is explicitly hidden from JSON serialization.

---

## 🟡 MINOR: Refactoring & Hardening
**Files:** `LoginRequest.kt`, `LogoutRequest.kt`

*   **Validation:** It would be cleaner to move the `require()` logic from the service to JSR-303 annotations (`@NotBlank`) on the DTOs. It’s more idiomatic and allows Spring to handle 400 errors automatically at the controller level.
*   **Rate Limiting:** We’re currently unprotected against brute-force attacks. Please open a follow-up ticket for implementing a throttling mechanism or account lockout policy.

---

## ✅ Code Quality Wins

*   **Logic:** Excellent job on username normalization and the idempotent logout implementation.
*   **Transactions:** Proper use of `@Transactional` across the service layer.
*   **Testing:** The test suite is very well-structured, particularly the assertions ensuring that "User Not Found" and "Wrong Password" return identical responses.

**Summary:** Address the side-channel in the service and the serialization leak in the domain model, and we’re good to go.

FabianSolutions:
Thanks for the thorough review @valentinaschiavon99! Pushed two fix commits addressing the blocking items, and filed follow-ups for the minor ones.

##  Critical — Timing side-channel fixed
**Commit:** `0479226` (`fix: #158 close timing side-channel in login`)

`AuthService.login` now always runs `passwordEncoder.matches(...)` — against the user's real hash if they exist, against a pre-computed `DUMMY_BCRYPT_HASH` constant otherwise. All three failure paths (unknown user, user with null `passwordHash`, wrong password) now take the same ~100ms.

The boolean gate is `user != null && user.passwordHash != null && passwordOk`, so the dummy-hash path can never be a successful login regardless of what plaintext the attacker sends.

**Test changes:**
- Renamed and flipped the existing `login rejects user with null passwordHash` test — encoder is now invoked (against the dummy), assertion changes from `verify(passwordEncoder, never()).matches(...)` to `verify(passwordEncoder).matches(eq, any)`.
- Added a new test `login runs BCrypt against a dummy hash when user is unknown for timing parity` that explicitly locks down the new contract so a future refactor can't silently re-introduce the side-channel.

##  Medium — `sessionToken` `@JsonIgnore` added
**Commit:** `f0cfb6f` (`fix: #158 mark UserModel.sessionToken as JsonIgnore`)

One-line change. No current endpoint serializes `UserModel`, but this closes the foot-gun ahead of #159.

##  Minor — Filed as follow-up issues
- **Rate limiting / lockout on `/auth/login`** → #169
- **JSR-303 `@NotBlank` migration** → #170

I deferred the JSR-303 migration because changing `require(...)` → annotation-driven validation also changes the 400 response shape (Spring's default JSON error body vs the current plain-text), and the Android client (Client #88, #92) parses plain text. To preserve the contract we'd need a `@ControllerAdvice` for `MethodArgumentNotValidException` — enough surface area that I think it deserves its own PR. Details in #170.

Ready for re-review.

Moritz2k1:
Well, this is highly overengineered for a uni project, but it is what it is. Good that we have rate limiting now which was originally mentioned in the out of scope section.

Moritz2k1 (approved):

Issue #170, opened 2026-05-01
Move auth DTO validation from require() to JSR-303 annotations


FabianSolutions:
Follow-up from #167 review (@valentinaschiavon99).

### Goal

Today `AuthService.register` / `login` / `logout` validate input with imperative `require(...)` calls. The reviewer suggested a more idiomatic approach: declarative `@NotBlank` annotations on the DTOs themselves, with `@Valid` on the controller's `@RequestBody` parameters so Spring rejects malformed requests before they reach the service.

### Why deferred from #167

Switching to JSR-303 changes the 400 response shape — Spring's default validation handler returns a JSON body with field-level details, while the current contract is a plain-text body containing the violated message. The Android client (Client #88, #92) parses the plain text. To preserve the contract we'd need a `@ControllerAdvice` for `MethodArgumentNotValidException` that flattens to plain text. That's enough surface area to warrant its own PR.

### Tasks

- [ ] Add `@NotBlank` (and length where the service currently checks `length <= 50` / `length <= 72`) to `RegisterRequest`, `LoginRequest`, `LogoutRequest`.
- [ ] Add `@Valid` to the controller params.
- [ ] Decide on the response shape:
  - (a) Accept Spring's default JSON error body and update the Android client to parse it.
  - (b) Add a `@ControllerAdvice` exception handler that converts `MethodArgumentNotValidException` to the existing plain-text format. Recommended.
- [ ] Remove the now-redundant `require(...)` calls from `AuthService`.
- [ ] Update controller + service tests to reflect the new entry point for validation errors.

### Out of scope

- `@ControllerAdvice` for the broader `Exception` cleanup mentioned in #162's PR description (separate cross-cutting issue).

### Related

Parent: #157
Source: PR #167 review (filed as a minor style / hardening suggestion, deferred to keep #167 scope tight).

eles17:
@FabianSolutions do we really need this?
@valentinaschiavon99

Issue #169, opened 2026-05-01
Add rate limiting / lockout to /auth/login


FabianSolutions:
Follow-up from #167 review (@valentinaschiavon99).

### Goal

Currently `/auth/login` has no protection against brute-force attempts. An attacker can hammer the endpoint with username + password guesses without any throttling, lockout, or backoff. Now that the timing side-channel from #167 is closed (BCrypt always runs), per-attempt cost is ~100ms — but a bot can still try thousands of passwords against a known username.

### Tasks

- [ ] Decide on the mechanism: per-IP rate limiting (e.g. token bucket via Bucket4j or Spring Cloud Gateway), per-username lockout (e.g. N failed attempts → temporary lock), or both.
- [ ] Implement and apply to `POST /auth/login` only (registration + logout don't have the same threat profile but worth considering).
- [ ] Decide failure response: 429 Too Many Requests vs the existing 401 with a generic body. The trade-off is "keep no-enumeration property" (401 always) vs "tell legitimate users to back off" (429 when locked).
- [ ] Decide whether locks unlock after a cooldown (e.g. 15 min) or require admin action.
- [ ] Tests covering: N failed attempts trigger throttling, throttle resets after cooldown, throttling is per-username (or per-IP) not global.

### Out of scope

- CAPTCHA (different threat model).
- Account-takeover monitoring / anomaly detection.
- Password-strength upgrades.

### Related

Parent: #157
Source: PR #167 review (timing fix shipped, throttling deferred).

eles17:
@valentinaschiavon99 says over-engineering. 

explain or we should close it - @FabianSolutions please close as not planned

valentinaschiavon99:
I believed it would have been a good thing to implement but seeing the status quo it's better to focus on other stuff and yeah probably it's over-engineering

eles17:
ok - otherwise we reopen later by need

Issue #104, opened 2026-04-19
Reconnection Logic for Active Games


valentinaschiavon99:
**Goal**
Ensure players can reconnect to the correct game state if they drop during or immediately after the start.
**Tasks**
[ ] Implement logic to map a reconnecting player's ID to their active IN_PROGRESS session.

[ ] Create a "Sync" event that sends the current game state to a re-joining player.

[ ] Handle the edge case where a player disconnects during the transition broadcast.
**Definition of Done**
[ ] Feature works as expected: A disconnected player returns to the game, not a lobby.

[ ] Edge cases handled: Handle reconnection when the server is mid-broadcast.

[ ] Tested manually: Refreshing the browser after the game starts puts the player back in the game.

[ ] Works in Docker environment.

[ ] PR created and CI passes.

PR #179, opened 2026-05-01
feat: #159 authenticate WebSocket sessions and attach Principal

### LINKED ISSUE #159: Authenticate WebSocket sessions and attach Principal

FabianSolutions:
Closes #159. Builds on #167 (login endpoint, already on `main`); unblocks Client #104 (send token in WS CONNECT) and Server #160 (active-player guard).

## Summary
- New `StompAuthChannelInterceptor` (`@Component`, registered on the inbound channel). Validates the session token on every STOMP `CONNECT` frame, looks up the user via `UserDao.findBySessionToken`, attaches a `UserPrincipal` to the WS session via `accessor.user`, rejects missing or invalid tokens.
- New `UserPrincipal` data class implementing `java.security.Principal` with `userId` + `username`.
- New `InvalidSessionTokenException` for the WS rejection path. Distinct from #167's `InvalidCredentialsException` — different code path, clearer logs.
- One-line wiring in `WebSocketConfig.configureClientInboundChannel`.

## Header contract
- **`Authorization: Bearer <token>`** as a STOMP native header on the CONNECT frame (`accessor.getFirstNativeHeader("Authorization")`).
- **Strict prefix** — bare tokens without `Bearer ` are rejected as malformed. Documented in code comments + the rejection test.

## No-enumeration / log hygiene
- Both rejection paths throw `InvalidSessionTokenException("Authentication failed")` with the **same** generic external message, so the STOMP ERROR frame body cannot be used to distinguish "no header" from "bad token". Internal WARN logs keep the distinction at the server for debugging.
- Token value is **never** logged (success path logs username only; rejection logs a generic reason).
- Username is sanitized for log injection (CR/LF stripped) before logging — belt-and-braces against malicious usernames. Charset validation at the registration layer is the real fix and is tracked under #170.

## What's intentional
- **Strict reject — no anonymous WebSocket sessions.** The issue body says "reject the CONNECT frame if missing or invalid" and that's what this does. Implication for clients: any pre-existing or unauthenticated client that connects on app startup will be rejected at CONNECT.
- **Existing handlers (`WebSocketController.addUser`, `GameController`) still pull usernames from the payload** — explicitly out of scope per the issue body's "out of scope" section. Future PRs can switch to `principal.name` when convenient.
- **Message rebuild** in the interceptor (`MessageBuilder.createMessage(payload, accessor.messageHeaders)`) instead of returning the original message after mutation. Both are valid per the `ChannelInterceptor` contract; the rebuild works reliably in both unit tests and production. The canonical Spring-docs pattern of "mutate accessor + return original message" depends on Spring's STOMP transport setting mutable headers, which is hard to recreate faithfully in a unit test. One extra `Message` allocation per CONNECT, negligible against the BCrypt-grade cost upstream.

## Out of scope (per issue body)
- **Issuing the token** — done in #158/#167.
- **Per-endpoint authorisation** (`principal.userId == activePlayer.userId`) — Server #160.
- **Switching existing handlers to use `Principal`** — flagged as nice-to-have follow-up, not required.
- **Token expiration / rotation** — Server #170 / future.
- **Rate limiting on CONNECT failures** — Server #169 / future.
- **Custom `StompSubProtocolErrorHandler`** for prettier client-facing ERROR frames — handled at the source by unifying the exception message; a fully custom error handler is a future polish item.

## Test plan
- [x] `./gradlew clean compileKotlin compileTestKotlin --warning-mode all` — clean, no warnings, no deprecation notices.
- [x] `./gradlew test --tests "org.machikoro.server.auth.*" --tests "org.machikoro.server.config.WebSocketConfigTests"` — all pass: 8 interceptor tests, 1 UserPrincipal test, 3 WebSocketConfig tests (including the new "interceptor registered" assertion).
- [x] `./gradlew test --rerun-tasks` — full suite green.
- [x] **Manual end-to-end** (recommended by reviewer) — the interceptor is invoked through Spring's actual STOMP routing; verify with:
  - `./gradlew bootRun` from this branch.
  - Connect via websocat or the Android client without an `Authorization` header → STOMP ERROR frame, server log `WS CONNECT rejected: missing or malformed Authorization header`.
  - Connect with `Authorization: Bearer not-a-real-token` → STOMP ERROR, server log `WS CONNECT rejected: token not recognised`.
  - Register + login + use returned `sessionToken` in `Authorization: Bearer <uuid>` → STOMP CONNECTED, server log `WS CONNECT accepted for user 'alice'`.
  - Could add a temporary debug `principal: Principal?` parameter to `WebSocketController.addUser` to confirm the principal reaches handlers, then remove before merge.

## Heads-up for reviewers
- **No integration test against Spring's real STOMP routing** — unit tests directly exercise `interceptor.preSend(...)` against mocked `UserDao`. To prove "Spring actually invokes my interceptor on CONNECT and respects the rejection" we'd need `@SpringBootTest` infrastructure with a real STOMP client. Manual E2E above is the practical equivalent for SE2 scope.
- **Strict-reject breaks unauthenticated clients.** Once this lands, the existing Android client (which connects on `MainActivity.onStart()` regardless of login state) will fail at CONNECT. This is **expected** and the counterpart fix is Client #104 ("defer WebSocket connect until session is non-null"). Coordinating these two merges is the deployment plan.

## Commit split (2)
1. `feat: #159 add UserPrincipal and InvalidSessionTokenException` — pure scaffolding.
2. `feat: #159 add StompAuthChannelInterceptor and wire into WebSocketConfig` — interceptor + 8 tests + WebSocketConfig wiring + the new config-level test.


Moritz2k1 (approved):
Adds a StompInterceptor for the sessionToken and the UserPrincipal

PR #187, opened 2026-05-02
refactor: extract EndTurnOutcome into separate data class

doniliks:
## Summary
Moved `EndTurnOutcome` from the end of the parent class into a standalone data class.

## Changes
- Created `EndTurnOutcome` data class
- Updated references to use the new class

## Notes
No functional changes intended.


valentinaschiavon99 (approved):
•
GamePhaseService.endTurn now returns a sealed interface EndTurnOutcome, which can either be Continue (the game proceeds to the next turn) or Won (a player has won).
•
GameController.endTurn now simply calls the service and, based on the outcome, either broadcasts the next game phase or broadcasts the winner.
**Potential Issues & Bugs:**
I don't see any obvious bugs. The logic seems sound and the refactoring improves the separation of concerns. The use of a sealed interface (EndTurnOutcome) for the return type of endTurn is a good practice, making the code in the controller cleaner and less prone to errors.
The transactionality in endTurn in GamePhaseService ensures that winner detection, game state updates, and cleanup happen atomically, which is great for data consistency.
**Conclusion:**
From my review of the code, I do not see any bugs. The changes represent a solid refactoring that improves the codebase.

PR #189, opened 2026-05-02
Feature/session recovery logic

### LINKED ISSUE #104: Reconnection Logic for Active Games

valentinaschiavon99:
This pull request introduces a robust reconnect and game state synchronization mechanism for players in active games. The main focus is ensuring that reconnecting users are reliably mapped back to their in-progress games and immediately receive an up-to-date game state snapshot, even in the presence of race conditions during game transitions. The changes include a new service for game state sync, enhancements to the WebSocket controller, supporting data access methods, and comprehensive tests.

**Reconnect and Game State Synchronization Improvements:**

* Added a new `GameSyncService` to encapsulate logic for finding a user's active in-progress game, checking membership, and building a complete game state snapshot for syncing reconnecting players.
* Enhanced `WebSocketController` to:
  * Prefer mapping reconnecting users to their server-tracked active game, handling transition races, and sending a SYNC message with the current game state to the user.
  * Introduce an explicit `/game.sync` endpoint for clients to request a state sync if they missed a game start event, with appropriate validation and state publishing.
  * Inject new dependencies: `GameSyncService` and `SimpMessagingTemplate` for message delivery. [[1]](diffhunk://#diff-1d1b7da08123cc3b5e3fece492bdb23f61c81e7446f51636936d90e564fe2befR4-R16) [[2]](diffhunk://#diff-1d1b7da08123cc3b5e3fece492bdb23f61c81e7446f51636936d90e564fe2befR25-R26)

**Data Layer Enhancements:**

* Added `findByGameIdAndUserId` and `findActiveGameIdByUserId` methods to `PlayerDao` to support efficient lookup of a user's current game and prevent duplicate player entries on reconnect.
* Updated `LobbyService` to avoid re-inserting players who are already in a game (reconnect path) and to protect against concurrent duplicate inserts.

**Protocol and DTO Updates:**

* Introduced `SyncGameRequest` DTO for explicit sync requests.
* Added `SYNC` as a new message type in `MessageType` to represent state sync events.
* Updated message type tests to include the new `SYNC` type.

**Testing Improvements:**

* Added comprehensive tests to `WebSocketControllerTests` covering:
  * Reconnect mapping and sync message delivery
  * Handling of race conditions during game start
  * Explicit sync endpoint behavior for various scenarios (valid, unauthorized, missing session, etc.)

These changes collectively ensure that reconnecting players experience a seamless transition back into their games, with immediate and accurate state synchronization.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -29,6 +29,7 @@ class LobbyWebSocketController(
      */
     @MessageMapping("/lobby.create")
     @SendTo("/topic/public")
+    @Suppress("UNUSED_PARAMETER") // Spring requires a @Payload parameter to deserialize the STOMP frame body
     fun createLobby(
         @Payload message: WebSocketMessage,
         headerAccessor: SimpMessageHeaderAccessor,
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -1,6 +1,6 @@
 package org.machikoro.server.controller
 
-import org.machikoro.server.dao.UserDao
+import org.machikoro.server.auth.UserPrincipal
 import org.machikoro.server.dto.MessageType
 import org.machikoro.server.dto.WebSocketMessage
 import org.machikoro.server.service.LobbyService
@@ -8,12 +8,12 @@ import org.slf4j.LoggerFactory
 import org.springframework.messaging.handler.annotation.MessageMapping
 import org.springframework.messaging.handler.annotation.Payload
 import org.springframework.messaging.handler.annotation.SendTo
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor
 import org.springframework.stereotype.Controller
 
 @Controller
 class LobbyWebSocketController(
     private val lobbyService: LobbyService,
-    private val userDao: UserDao
 ) {
     private val logger = LoggerFactory.getLogger(LobbyWebSocketController::class.java)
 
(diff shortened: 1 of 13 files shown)


FabianSolutions (changes requested):
1. Username spoofing → state exfiltration. addUser still reads identity from message.sender (client-supplied), and the new path uses that identity to push the full GameStateDto snapshot to the caller. So an unauthenticated client can send JOIN { sender: "alice" } and receive Alice's active game state. The known #159 gap on addUser is now actively exploitable through the SYNC channel. Please switch to the UserPrincipal that StompAuthChannelInterceptor attaches in #159 — read identity from headerAccessor.user as? UserPrincipal rather than from the payload. (#160 just landed and uses this pattern for the game-mutating endpoints — same approach here.)
 
2. convertAndSendToUser(sessionId, ...) won't route in production. The first argument to convertAndSendToUser is the Principal name, not a raw STOMP sessionId — Spring's UserDestinationResolver looks up the user's session by Principal. Passing a sessionId means the message is silently dropped. The unit tests don't catch this because SimpMessagingTemplate is mocked. Either pass principal.name (once you do fix 1) or test against a real broker. Right now the SYNC delivery is unverified.

Two smaller follow-ups that are fine to file separately and not block this PR on:

 - N+1 in buildSnapshot — one findByPlayerId per player. Tiny lobbies today, but a bulk DAO method would be cleaner. Worth a follow-up issue.
- The GameStartedException catch falls back to findActiveInProgressGameId(user.id) which may resolve to a different game than the client asked for. The recovery semantics deserve a comment explaining when this is intentional
vs. a sign of an inconsistent client.

Once 1 and 2 are addressed I can re-review.

FabianSolutions (approved):
Both blockers are properly resolved. LGTM, approving.

PR #194, opened 2026-05-04
Feature/add new establishment type enum

### LINKED ISSUE #192: Add EstablishmentType enum for card effects and UI icons
## Introduce an EstablishmentType enum to represent categories of establishments. This will be used for:

- Card effect logic (e.g., “gain coins per RESOURCE card”)
- Mapping UI icons consistently across all cards

Acceptance Criteria:

- [x] EstablishmentType enum is added to code and database
- [x] Each card has an assigned EstablishmentType (#182)
- [ ] Card effects can reference types in  #181 (e.g., count cards by type)

doniliks:
This PR introduces a missing `EstablishmentType` enum to categorize establishments and assigns each CardType a corresponding type.

This enum was missing in the actual DB implementation. As planed, this enum is now explicitly added to support both backend and UI.

The goal is to:

- Enable game effect logic (e.g., “gain coins per RESOURCE card”)
- Provide a consistent mapping for UI icons per establishment type

---

Changes:

* Added new enum: `EstablishmentType`
* Extended `CardType` to include `establishmentType`
* Assigned all currently existing cards an appropriate `EstablishmentType`
* Updated tests to validate the new structure
* Adjusted existing test accordingly

Type | Original Icon
-- | --
AGRICULTURE | Fruit/Wheat
LIVESTOCK | Cow
RESOURCE | Gear
RETAIL | Bread
INDUSTRY | Factory
HOSPITALITY | Cup
MAJOR | Purple major establishments
MARITIME  | Boat
ENTERPRISE  | Suitcase 

Hospitality icon can be changed to add hotels

---

Notes:

* Due to differences in DB design, it is still TBD whether:
  - landmarks will use `EstablishmentType.LANDMARK`, or
  - landmarks will be handled separately and rendered differently in the UI
  - some landmarks also include effects that interact with cards by EstablishmentType, which may require reconsidering the DB layout adding inheritance, change card type etc. (TBD by the responsible for card effects).
* Additional types like `MARITIME`, `ENTERPRISE`, added for variety of cards, since its planned to add more than 15 Cards for Marketplace, and `LANDMARK` is included for future expansion


Moritz2k1:
Maritime and Enterprise are from the Machi Koro expansion and do not belong here.
I don't see the value in adding an establishment type other than grouping the icons for the frontend. Right now we can do the whole game logic with the CardType and CardColor. Other than that the EstablishmentType does belong into the table and not on the CardType (own Column in DB)

FabianSolutions:
No value in base machi koro. Would be worth if a group had more cards or was referenced by multiple effects which is not the case in base machi koro. And so far we did not plan to implement any expansion where it would add some value.

doniliks:
"Maritime and Enterprise are from the Machi Koro expansion and do not belong here."

I added them just to have different options for card names since having 12 versions of farm, ranch, farmer house, backyard etc. will be not interesting to play with.


"I don't see the value in adding an establishment type other than grouping the icons for the frontend. " 

Some cards effect generate income based on small icons all cards have, e.g "Get 1 coin for each "Cow" and "Wheat" icon. In UML i described them as EstablishmentType to iterate through cards based on it and get value of this addtional income. Currently we have no working effects, but this can be implemented now.

Also we still will have to render all cards during whole game, so i assume it would be better to save it directly. The same story for Landmarks. In UML it has a generalization (Java-Style with abstract class) Card is-a Landmark and Establishment to keep all relevant info such as name, description, icon, type, is triggered etc. and store additionally effects and additional flags in instance. Here we could make one of those 3 ways of generalisation so we will pass same values for all cards to render. Also adding one cardId for all cards should make buying phase not that complex.
I assume this setup would make sense but haven't implemented something like this b4.

"Right now we can do the whole game logic with the CardType and CardColor" 

Yes we can make this basic calculation now, but we will still have to expand it in future, so why not adding it now.


"EstablishmentType does belong into the table and not on the CardType (own Column in DB)" 

I made it like this at first, but then i saw in code that we refer to color from CardType. Since Color is currently not saved in extra column, i saved this value in CardType as well. 
Here we should reconsider this and move it in general to separate columns, since I don't like idea of having CardType as enum with card names in general. Maybe we could have just unique names of cards in DB and all other values as well.
Same applies for others, where we could get rid of enums and save all in DB only.

doniliks:
image
 Also this appears in base game, thats why i added it to uml at the beginning

valentinaschiavon99:
I have reviewed this PR and discussed the implementation details with @doniliks .
 From an architectural standpoint, the refactoring of the communication layer and the logic updates make sense for the long-term scalability of the server.

However, we have a significant blocker regarding the data layer: the database schema must be changed.

Let's discuss this during this Thursday's meeting to decide on the best course of action and ensure we don't break the dev environment.

Moritz2k1:
> "Maritime and Enterprise are from the Machi Koro expansion and do not belong here."
> 
> I added them just to have different options for card names since having 12 versions of farm, ranch, farmer house, backyard etc. will be not interesting to play with.
> 
> "I don't see the value in adding an establishment type other than grouping the icons for the frontend. "
> 
> Some cards effect generate income based on small icons all cards have, e.g "Get 1 coin for each "Cow" and "Wheat" icon. In UML i described them as EstablishmentType to iterate through cards based on it and get value of this addtional income. Currently we have no working effects, but this can be implemented now.
> 
> Also we still will have to render all cards during whole game, so i assume it would be better to save it directly. The same story for Landmarks. In UML it has a generalization (Java-Style with abstract class) Card is-a Landmark and Establishment to keep all relevant info such as name, description, icon, type, is triggered etc. and store additionally effects and additional flags in instance. Here we could make one of those 3 ways of generalisation so we will pass same values for all cards to render. Also adding one cardId for all cards should make buying phase not that complex. I assume this setup would make sense but haven't implemented something like this b4.
> 
> "Right now we can do the whole game logic with the CardType and CardColor"
> 
> Yes we can make this basic calculation now, but we will still have to expand it in future, so why not adding it now.
> 
> "EstablishmentType does belong into the table and not on the CardType (own Column in DB)"
> 
> I made it like this at first, but then i saw in code that we refer to color from CardType. Since Color is currently not saved in extra column, i saved this value in CardType as well. Here we should reconsider this and move it in general to separate columns, since I don't like idea of having CardType as enum with card names in general. Maybe we could have just unique names of cards in DB and all other values as well. Same applies for others, where we could get rid of enums and save all in DB only.

1. I literally have no idea what you mean. We are implementing the base game of machi koro. If you want to implement the expansions, feel free to fork the repo. Also there are no backyard or farmer house cards. I don't understand the sentence.
2. There are literally no cow icons. Other than that, you can still do the whole shopping mall logic for example by just referencing the cardTypes. I don't understand the issue, but I guess this would be a valid point for an establishmentType
3. Frontend rendering is a frontend concern - the frontend can merge two response DTOs itself. We don't need to restructure the DB for this. The single cardId idea adds complexity without clear benefir for the base Machi Koro game - the buying phase already knows whether you're buying an establishment or a landmark since they come from different places in the UI. No need to over-engineer.
4. I can move the CardColor to be in the table as well and we can change from an enum approach to a full db column approach if wanted

doniliks:
1. Once again, we discussed that we want to add different cards to enhance variety of round outcome. That means we will add more cards than in original game and Maritime and Enterprise are just new icons on cards that allow to expand current Establishments with just new NAMES on cards. Adding new cards easily should be not a problem for us in our project.

2. "Also there are no backyard or farmer house cards." Those are just random cards showing that currently we are limited in variation of card names. Also I mentioned that I called "Cup" for Restaurants as Hospitality since we could add hotels just idk "Hotel - player pays 5 coins if rolled this dice" same way as restaurant. This is just new card for visual part of game to let players see different outcome not just 3 restaurants from original game, it will be boring otherwise.
Cow is a icon on ranch card, names are from wiki for game. 

3. "Also there are no backyard or farmer house cards. I don't understand the sentence." because in Agriculture we have limited options and they are presented in game, adding synonymes for farm is just boring. 
I don't really understand why do you consider expansion of original game like it introduces something completely new, those are just cards with some new effects, and we are free to choose them on our own and customise it.

And also at this point of time we have nothing that renders UI, not even cards, so yes, maybe with this current setup we could somehow manage it, but like already implemented part with buying phase has problems because of DB and was also not testet with real front by real actions. 

4. I cant really comprehend how to find a number of all cards of specific type if we currently having a name and colour., thats it. Adding this basic enums allows to iterate and find those cards faster and makes adding new cards easier, since idk add a list of card that this shopping mall activates, this is really not expandable staff in general. If you will look at the card as a class, this icon appears at every card even at landmarks so saving it this way in a normal solution for both back and front.
 
5. We also discussed in person and got opinion from AI that current DB has to changed. 

Independent AI opinion on what has to do with card, other tables have some problems as well.

"Using cardId instead of separate CardType/LandmarkType is better because:

one unified purchase flow
fewer DTO fields
no nullable combinations
simpler WebSocket messages
easier frontend handling
easier DB queries
easier expansions/new cards
less branching logic in services
game becomes data-driven instead of enum-driven

Basically:
cardId = one source of truth
instead of duplicating card identity in enums + code logic."

and 

"EstablishmentType is useful because some cards depend on categories of other cards.

Example in Machi Koro:
Cheese Factory → earns per COW
Furniture Factory → earns per GEAR
Fruit Market → earns per FRUIT

Without EstablishmentType, you end up hardcoding lists like:
if (cardType == RANCH || cardType == CHEESE_FACTORY)
With it:
countCardsByType(EstablishmentType.COW)"

and last 

"Current:

data class PurchaseRequest(
    val gameId: Int,
    val purchaseType: PurchaseType,
    val cardType: CardType? = null,
    val landmarkType: LandmarkType? = null,
)

Needs:
resolvePurchaseTarget(...)
and:
when (purchaseType)

Refactored:

data class PurchaseRequest(
    val gameId: Int,
    val cardId: Int,
)

Then:

val card = cardDao.findById(cardId)

when (card.category) {
    ESTABLISHMENT -> ...
    LANDMARK -> ...
}

No nullable fields.
No resolver.
No invalid combinations."

So we have to change it asap because now we already have to refactor existing code to adjust it, but we already had to be ready with real effects resolution at this point of time. 

Otherwise we will refactor code longer than it actually was coded, what happens now e.g. removing/adding some parts by different contributors because we stuck somewhere and cant move further after this basic enum PR revealed more problems that it could be.  Lets discuss it tomorrow to have other opinions from everyone, approve final design version and see how we will resolve it.

Moritz2k1:
Since the other PR (refactor database) has been merged, I will close this one.

PR #217, opened 2026-05-09
fix: persist websocket principal for lobby creation

Ivona-Goranova:
Fixed WebSocket authentication propagation for lobby creation.

The STOMP CONNECT frame was authenticated successfully and a UserPrincipal was attached to the accessor, but the authenticated user was not available in later @MessageMapping handlers (e.g. /lobby.create). As a result, lobby creation failed with "Authenticated principal not found" and no LOBBY_CREATED event was sent back to the client.

Changes:

* Persisted the authenticated UserPrincipal in sessionAttributes inside StompAuthChannelInterceptor
* Updated LobbyWebSocketController to resolve the principal from either accessor.user or sessionAttributes
* Verified that the lobby creation flow now correctly broadcasts LOBBY_CREATED messages and displays the generated lobby code on the client

Tested manually:

* Login via Android client
* Successful WebSocket CONNECT authentication
* Lobby creation via /app/lobby.create
* Reception of LOBBY_CREATED message
* Lobby code displayed correctly in UI


FabianSolutions (dismissed):
The sessionAttributes approach was correct, but only patched one of three affected sites.

  Diagnostic on a live server + Android run confirmed accessor.user is null on every SEND frame in our setup (Spring's per-frame principal propagation needs either WebSocketSession.principal from the HTTP handshake or Spring
  Security's WebSocket integration; we have neither because the token is in the STOMP CONNECT body). Session attributes do propagate, which is why this PR's workaround works.

  Followup commit 0507b0f extracts a SimpMessageHeaderAccessor.userPrincipal() helper, applies it at all three affected sites (lobby.create, chat.addUser, game.sync — chat.addUser was already silently broken before this PR),
  switches the lobby RuntimeException to CustomWebSocketException("UNAUTHENTICATED", ...) so it routes via GlobalWebSocketExceptionHandler, and adds matching tests.

  Filed #218 for the GameController endpoints — likely affected by the same root cause but kept out of this PR to scope it tight. Approving.

FabianSolutions:
Can someone please review and merge? Since I did some changes here I can't review and merge.

Moritz2k1 (approved):
Looks fine to me. Implements persistency for lobby creation. Has been quickly adjusted by Fabi, which is fine.

Moritz2k1 (approved):

FabianSolutions (approved):
Merging

Issue #46, opened 2026-04-03
Phase Changes


SoupCube:
### User Story
As a player I want the game to be separated into different stages, that automatically switch on certain triggers.

### Acceptance Criteria
- In the beginning, the rolling phase starts after the initialization
- During the, game the rolling phase starts after the buying phase of the previous player
- The earning phase starts after the rolling phase is completed
- The buying phase automatically starts after the earning phase has been completed
- Phase state is synchronized for all players

#### Related Issues
- https://github.com/SE2-Machi-Koro/Client/issues/17

Moritz2k1:
This issue can be closed since all of the sub-issues have been fulfilled

PR #221, opened 2026-05-11
117 store dice result in game state story

### LINKED ISSUE #117: Store Dice Result in Game State Story
As the server, I want to store the dice result in the current game state so that further game logic can access it.
Acceptance Criteria:

-     Result is saved in the game state object for the current turn
-     Result is accessible for subsequent actions (e.g. card activation)
-     Result is reset at the end of the turn

kutschera13:
## Description
As the server, I want to store the dice result in the current game state so that further game logic can access it, and broadcast the active player ID with each phase change so the client can restrict the dice roll button to the active player only.

## Acceptance Criteria
- [x] Result is saved in the game state object for the current turn
- [x] Result is accessible for subsequent actions (e.g. card activation)
- [x] Result is reset at the end of the turn
- [x] `activePlayerId` is included in every `GAME_ACTION` broadcast

## Implementation Notes
- `DiceService.rollDice()` calls `gameDao.updateAfterRoll()` to persist the result and advance to `RESOLVE_EFFECTS`
- `GameDao.advanceTurn()` resets `lastDiceRoll = null`
- `GameController.broadcastPhase()` now includes `activePlayerId` by looking up the current player via `playerDao.getPlayers(gameId)[currentTurnIndex].userId`

## Related
- Follows up on Client #46

## Closes
Closes #117


Moritz2k1:
Lea, can you please not remove the comments.

eles17 (dismissed):
I reviewed this PR separately from #222. As discussed on WhatsApp the #222 should be changed and remove duplicate work after this PR is merged.

But regarding this PR, the goal makes sense: storing the dice result in game state and sending the active player info with phase broadcasts is useful for the client. This should be merged before the follow-up dice broadcast PR.

A few things should be fixed/clarified before merge:

1. One small naming question: `activePlayerId` currently sends `activePlayer.userId`, not `activePlayer.id`.
- If the client should compare this value with the logged-in user ID, then the value is correct, but maybe the field should be called `activeUserId` to avoid confusion.
- If we want to keep the name `activePlayerId`, then I would expect it to contain the actual player row ID (`activePlayer.id`). Could you clarify which ID the client should use here?


2. Please restore the removed comments.
   This PR removes several existing controller comments that explain the WebSocket handlers. That is unrelated to the story and was already mentioned in the PR conversation.

3. Remove temporary comments like `// NEU`.
   These should not stay in final merged code.

4. Please resolve open conflicts. 

From my side, the story itself looks fine, but I would clean up the naming/comments and resolve the merge conflict before merging.

Moritz2k1 (approved):
Dice result is getting stored in the payload now, comments have been restored

PR #227, opened 2026-05-11
Fix/195 critical bugs

### LINKED ISSUE #196: Critical - Red Card Logic
Problem: Red card effect is deducting coins from the bank instead of the active player.
Impact: Breaks core game mechanics and economy balance.
Action: Fix logic to correctly transfer coins from the active player.

### LINKED ISSUE #197: Critical - Race Condition in addUserToLobby
Problem: Status check occurs outside of the lock.
Impact: Can lead to inconsistent lobby state or duplicate joins.
Action: Move status check inside the lock to ensure atomicity.

### LINKED ISSUE #198: Critical - resolveEffects Transaction Split
Problem: Uses two separate transactions.
Impact: Risk of double coin distribution or partial updates.
Action: Refactor into a single transaction or ensure proper transactional guarantees.

Moritz2k1:
Fixed critical bugs that were highlighted in the following issues:

- [x] resolveEffects Transaction Split
- [x] Race Condition in addUserToLobby
- [x] Red Card Logic


eles17 (changes requested):
I checked this PR against the three linked issues.

`resolveEffects Transaction Split` looks addressed: earnings and the phase transition now happen inside one transaction, which avoids partial state.

`Race Condition in addUserToLobby` also looks addressed: the game/player state is re-checked inside the lock before inserting the player.

For `Red Card Logic`, the change goes in the right direction because red cards now subtract from the active player instead of paying from the bank. But I think the transfer amount still needs one correction before merge:

If the active player does not have enough coins, the red-card owner should only receive the amount the active player can actually pay.

Example:
- active player has 0 coins
- Cafe wants to take 1 coin

Current behavior:
- active player stays at 0 because of clamping
- Cafe owner still gets +1

That creates a coin instead of transferring one. The amount given to the Cafe owner should be limited by how many coins the active player actually has..

Same idea probably applies to `ALL_PLAYERS` payments like Stadium: each opponent should only pay up to the amount they actually have, and the card owner should receive the sum of the actual payments.

Buy/Build itself is separate from this PR. I will handle the Buy/Build phase refactor in my next branch #228 .

Moritz2k1:
Request changes have been implemented

eles17 (dismissed):
Thanks for the work here. The test suite passes locally (`./gradlew test`), but I’m requesting changes before merge because there are two remaining correctness issues:

- `addUserToLobby` now protects the join flow with a per-lobby lock, but `startGame` does not participate in that same lock. That means a game can still move to `IN_PROGRESS` between the join flow’s status check and player insertion.
- `PaymentSource.ALL_PLAYERS` currently divides the card income across opponents. For Stadium, each other player should pay the card income amount, capped by their available coins; the current logic underpays in 3-4 player games.

Once those are fixed and covered by tests, this should be in good shape.

Moritz2k1:
Requested changes have been implemented again

FabianSolutions (dismissed):
LGTM. Problems were addressed. Merged

FabianSolutions (approved):
LGTM. Merging

PR #232, opened 2026-05-12
feat: Integrate Flyway for database migrations

### LINKED ISSUE #230: Add Flyway Database Migration Setup
## Goal

Add Flyway migration support so database schema changes are versioned and reproducible.

## Tasks

- [x] Add Flyway dependency
- [x] Configure Flyway for the database
- [x] Create initial migration scripts
- [x] Follow Flyway filename pattern, e.g. `V1__init_schema.sql`
- [x] Verify migrations run on application startup
- [x] Document basic migration usage if needed

## Definition of Done

- [x] Flyway is configured
- [x] Migration scripts follow required naming pattern
- [x] Database schema can be created through migrations
- [x] App starts with migrated database
- [x] PR created and CI passes
- [ ] Create a post on Moodle about it.

valentinaschiavon99:
This pull request introduces database schema management using Flyway and adds an initial migration for the `users` table. The key changes are as follows:

**Database migration setup:**

* Enabled Flyway for automatic database migrations and configured it to use migration scripts from the `classpath:db/migration` directory. Also set `baseline-on-migrate` to ensure compatibility with existing databases. (`application.properties`)

**Schema initialization:**

* Added an initial Flyway migration script (`V1__init_schema.sql`) that creates a `users` table with fields for user identification, authentication, and tracking game statistics. (`db/migration/V1__init_schema.sql`)…chema


eles17 (changes requested):
Not mergeable yet.

Findings


[P1] Flyway migration is incomplete: V1__init_schema.sql (line 1) only creates users, but the app schema has cards, card_activation_numbers, landmarks, games, players, player_cards, player_landmarks, and game_marketplace in Tables.kt (line 26). With machikoro.db.init.enabled=false, a fresh DB will be missing most tables.



[P1] Flyway and Exposed schema creation both remain active: application.properties (line 13) keeps machikoro.db.init.enabled=true, while application.properties (line 20) enables Flyway. That means Flyway is not actually the source of truth; Exposed still creates the real schema at startup. Future migrations can drift from the runtime schema silently.



Verification

./gradlew test passed.

./gradlew check passed after allowing Gradle wrapper access to its external lock file.

Verdict

The branch is buildable, but I would not merge it as a Flyway integration. It needs either a complete initial migration and Exposed schema auto-creation disabled, or the Flyway change should be scoped honestly as experimental/non-authoritative.

eles17 (approved):
I re-reviewed the updated Flyway integration and the previous blockers have been addressed: the initial migration now covers the full Exposed schema plus required reference data, and runtime Exposed schema initialization is disabled by default so Flyway becomes the authoritative schema source.

I also verified the relevant gates locally:

- `./gradlew check`
- `./gradlew dependencies --write-locks --quiet` with no generated diff
- `./gradlew clean build test --info --stacktrace`
- `./gradlew help --task sonar`

Issue #249, opened 2026-05-13
Reconnect integration test for /game.sync covering full snapshot


FabianSolutions:
## Scope

Add a DB-backed integration test (Testcontainers, sibling to `LobbyServiceIntegrationTest`) that exercises the full reconnect snapshot path:

1. Creates a game with 2 players.
2. Mutates state: one player buys an establishment, builds a landmark, rolls dice, advances phase.
3. Invokes `GameSyncService.buildSnapshot(gameId)`.
4. Asserts the returned `GameStateDto` correctly reflects **every** field:
   - `game.status`, `game.turnPhase`, `game.currentTurnIndex`, `game.lastDiceRoll`, `game.roundNumber`
   - both players' `coins`
   - both players' owned `playerCards`
   - both players' `playerLandmarks` (built vs unbuilt)
   - `marketplace` supply decremented by 1 for the bought card.

Locks down the contract so future regressions on any field surface immediately.

## Parent

#246

## Depends on

- Server #247 (playerLandmarks field) — assumed to be the next-issued number.
- Server #248 (marketplace field) — same.

Issue #246, opened 2026-05-13
Complete Game State Snapshot for Reconnect


FabianSolutions:
## Goal

Reconnect via `/app/game.sync` must return the full playable state so the client can re-render the game faithfully — game/phase, active player, players, coins, cards, landmarks, marketplace, dice result, round. Maps directly to the professor's Sprint 2 checklist item *"Game State Recovery bei Dis/Reconnect"*, which is currently only partially satisfied because the snapshot is incomplete.

## Current state

`GameStateDto` (returned by `GameSyncService.buildSnapshot`) today includes:

- `game` (status, turnPhase, currentTurnIndex, lastDiceRoll, roundNumber, hasPurchasedThisTurn) ✓
- `players` (coins, turnOrder, lastSeenAt) ✓
- `playerCards` (per-player owned establishments) ✓
- `turnOrder` (sorted list of player IDs) ✓

**Missing from the snapshot:**

- `playerLandmarks` — built/unbuilt status of each player's landmarks. Without this, the client can't render landmark progress on reconnect (and the player has no idea how close anyone is to winning).
- `marketplace` — remaining supply per card type. Without this, the client can't render what's still available to buy.

Screen destination is derivable on the client side from `game.status` + `game.turnPhase`, so no extra server field is needed for routing — the snapshot completeness alone is sufficient.

## Sub-tasks

- [x] #247 Add `playerLandmarks` to `GameStateDto` and `GameSyncService.buildSnapshot`
- [x] #248 Add `marketplace` supply to `GameStateDto` and `GameSyncService.buildSnapshot`
- [ ] #249 Reconnect integration test for `/game.sync` covering full snapshot
- [ ] SE2-Machi-Koro/Client#140 Render full reconnect snapshot on the client (landmarks + marketplace + full state)

## Acceptance

- Reconnect via `/app/game.sync` mid-game returns a payload from which the client can fully reconstruct: routing decision, current phase, whose turn it is, every player's coins, every player's owned cards, every player's landmark build state, what's still in the marketplace, the last dice roll, the round number.
- Backed by an integration test that creates a game, mutates state, calls sync, and asserts every field on the returned `GameStateDto`.
- Client renders the reconstructed state without any extra round-trips after a single `game.sync` response.

Issue #140, opened 2026-04-26
Move DAO/DSL explanations to README and remove redundant comments from codebase


valentinaschiavon99:
## Description
To improve code clarity and maintainability, we should:
- Remove redundant or generic comments from DAO, DSL, and related files.
- Move general explanations about what a DAO or DSL is to the README file, so that architectural concepts are documented in a central, accessible place.
- Ensure that only meaningful, context-specific comments remain in the code.

## To Do
- [x] Identify all DAO, DSL, and related files containing redundant or generic comments
- [x] Remove unnecessary comments from these files
- [x] Draft a clear explanation of DAO and DSL concepts for the backend
- [x] Add these explanations to the README file
- [x] Review the codebase to ensure only valuable comments remain
- [x] Share the updated README with the team for feedback

## Definition of Done
- All redundant or generic comments are removed from DAO, DSL, and related files
- README contains a clear, concise explanation of DAO and DSL concepts
- Only meaningful, context-specific comments remain in the codebase
- Team has reviewed and approved the updated documentation

_Originally posted by @Moritz2k1 in https://github.com/SE2-Machi-Koro/Server/issues/135#issuecomment-4321787821_

Moritz2k1:
I have removed the redundant comments in the DAOs and Entities files and implemented a description of what DAO's, DSL, the Entity API, Entities and DTOs are

PR #250, opened 2026-05-13
feat: include playerLandmarks in GameStateDto and sync snapshot

### LINKED ISSUE #247: Add playerLandmarks to GameStateDto and GameSyncService snapshot
## Scope

Extend `GameStateDto` with a player→landmarks map:

```kotlin
val playerLandmarks: Map<Int, List<PlayerLandmarkModel>>,
```

In `GameSyncService.buildSnapshot`, fetch via the existing `playerLandmarkDao.findByPlayerId(...)` per player (or a batch variant if added) and populate the new field — same pattern the existing `playerCards` map uses.

## Acceptance

- `GameStateDto.playerLandmarks` is populated on every snapshot.
- Test asserts the returned map has one entry per player and the entries reflect what `PlayerLandmarkDao` has stored, including built/unbuilt state.

## Parent

#246

FabianSolutions:
Closes #247. Part of #246 (Complete Game State Snapshot for Reconnect).

## Why

Reconnect via `/app/game.sync` currently returns players, coins, owned cards, and turn order — but not landmark build state. The client can't render the build grid on reconnect without an extra round-trip. The same gap exists for the initial \`GAME_STARTED\` broadcast.

## What changed

- **\`GameStateDto\`** gains a required \`playerLandmarks: Map<Int, List<PlayerLandmarkModel>>\` field, mirroring the existing \`playerCards\` shape. No default value, so every construction site is enforced at compile time. Swagger schema + KDoc updated to describe the dual purpose (game start broadcast + sync response).
- **\`GameSyncService.buildSnapshot\`** injects \`PlayerLandmarkDao\` and populates the new map. Per-player iteration is bounded at \`N = maxPlayers = 4\` — a comment points to where a batch DAO method would go if that ever matters.
- **\`LobbyService.startGame\`** includes the freshly-initialized landmarks in the initial snapshot — same transaction as \`initForPlayer\`, so reads see the just-written rows. Clients get the full build grid in one message instead of needing a follow-up sync.
- **\`PlayerLandmarkDao.findByPlayerId\`** adds \`.orderBy(Landmarks.landmarkType to SortOrder.ASC)\`. Stable ordering matters now that the result is serialized to clients — without it the JSON output (and any downstream tests, e.g. the integration test in #249) would be at the mercy of the DB's row-return order.

## Tests

- **\`GameSyncServiceTest\`** — added \`PlayerLandmarkDao\` mock; extended the primary snapshot test to stub one built and one unbuilt landmark and assert both round-trip per player; added \`playerLandmarks.isEmpty()\` assertion to the zero-players test.
- **\`LobbyServiceIntegrationTest.startGame ...\`** — now asserts the new field is populated for both players and that every landmark starts unbuilt.
- **\`GameControllerTest\` + \`LobbyControllerTests\`** — \`gameStateDto()\` helpers updated with \`playerLandmarks = emptyMap()\` (compile-forced since the field has no default).

## Wire-format note

This is an **additive** field on a JSON DTO. Existing clients that don't yet read \`playerLandmarks\` continue to work — Jackson's default deserializer ignores unknown-on-write fields. The matching client-side render work is tracked in [Client #140](https://github.com/SE2-Machi-Koro/Client/issues/140).

## Test plan

- [x] \`./gradlew compileKotlin compileTestKotlin --warning-mode=all\` — no warnings.
- [x] \`./gradlew test --tests <affected>\` — \`GameSyncServiceTest\`, \`LobbyServiceTest\`, \`GameControllerTest\`, \`LobbyControllerTests\` all green.
- [ ] Testcontainers integration tests (\`LobbyServiceIntegrationTest\`) — not run locally (Docker not available); will exercise in CI.
- [ ] Sonar runs in CI.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/dao/PlayerLandmarkDao.kt
+++ b/src/main/kotlin/org/machikoro/server/dao/PlayerLandmarkDao.kt
@@ -1,6 +1,7 @@
 package org.machikoro.server.dao
 
 import org.jetbrains.exposed.v1.core.ResultRow
+import org.jetbrains.exposed.v1.core.SortOrder
 import org.jetbrains.exposed.v1.core.and
 import org.jetbrains.exposed.v1.core.eq
 import org.jetbrains.exposed.v1.jdbc.deleteWhere
@@ -29,12 +30,17 @@ class PlayerLandmarkDao {
     )
 
     /**
-     * Finds all landmarks for a given player
+     * Finds all landmarks for a given player, ordered alphabetically by
+     * landmark type. Stable ordering matters because the result is surfaced
+     * in `GameStateDto.playerLandmarks` and serialized to clients — without
+     * an explicit order, the JSON output (and any downstream test assertions)
+     * would be at the mercy of the DB's row-return order.
      */
     fun findByPlayerId(playerId: Int): List<PlayerLandmarkModel> = transaction {
         (PlayerLandmarks innerJoin Landmarks)
             .selectAll()
             .where { PlayerLandmarks.playerId eq playerId }
+            .orderBy(Landmarks.landmarkType to SortOrder.ASC)
             .map { it.toPlayerLandmarkModel() }
     }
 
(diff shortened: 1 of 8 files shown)


Moritz2k1 (approved):
PlayerLandmarks get sorted now. GameStateDTO has been extended by playerLandMarks which makes sense. LobbyService and GameSyncService also got extended by the playerLandmarks

Issue #254, opened 2026-05-14
Verify Game Recovery After Backend Container Restart


FabianSolutions:
## Goal

Prove end-to-end that an in-progress game survives a backend container restart. Maps directly to the professor's Sprint 2 checklist item *\"Game state recovery upon failure/restart of the backend container\"* — this is an explicit grading requirement and gets checked live during the demo.

## Current state (what's already in place)

The recovery **mechanism** is built; what's missing is documented **verification**.

- Game state is fully persisted in Postgres via Flyway-managed schema (#230) — no in-memory game state survives across restarts; all of it is in the DB.
- `/app/game.sync` returns the full reconnect snapshot covering every state field after the #246 chain: game/phase/dice/round (#247, #248, #249).
- Clients reconnecting to an existing game route through `LobbyService.addUserToLobby`'s reconnect short-circuit, which returns the player's record without re-insert.
- Auth tokens are stored client-side (`SessionStateHolder`), so the WS reconnect after the container comes back up uses the same session.

In principle a `docker compose restart server` should not lose state. **We have no test that proves it.**

## Scope

1. **Documented manual test procedure** in the network protocol doc (or a separate `docs/backend-restart-recovery.md`) covering:
   - Start the backend via `docker compose up`.
   - Two clients log in, host creates a lobby, both join, host starts the game, players take a few turns (purchase a card, build a landmark, advance phase).
   - `docker compose restart server` (or `docker stop` + `docker start`).
   - Both clients reconnect.
   - Verify: state is identical to pre-restart — same active player, same coins, same owned cards, same landmark build state, same marketplace, same phase, same round.
   - Game continues normally from where it left off.

2. **At least one of**: an automated smoke test (shell script that drives the above flow) OR a Testcontainers-based integration test that actually restarts the Postgres-backed container and re-issues `/app/game.sync` against the restarted server. Whichever is cheaper to ship before the Sprint 2 due date (2026-05-19).

3. **A short evidence section** in the Sprint 2 PDF / submission referencing the test procedure + a screenshot or log excerpt showing the recovery actually working.

## Acceptance

- A reviewer following the documented procedure can restart the server container mid-game and resume play without re-creating the lobby or losing any state field.
- The procedure is referenced from the network protocol doc.
- Either a smoke test or an integration test exists and passes against a restarted container.
- The Sprint 2 submission cites this verification.

## Parent of (to be filed if scope splits)

[...]

PR #255, opened 2026-05-14
fix: re-enable Exposed bootstrap so DAO transactions have a database

FabianSolutions:
## What's broken

The Android client surfaces a raw server exception when you tap Register:

> No database specified and no default database found. Please call Database.connect() first or specify a database explicitly in the transaction

Every DB-touching endpoint on `main` is currently broken — register, login, lobby create, lobby join, gameplay, sync. The server boots fine; the failure only fires on the first DAO call. That's why nobody caught it in CI: integration tests override the relevant property and never exercise the production default.

## Root cause

Commit 443b37f (\"feat: Update database initialization settings for migration-driven setup\") flipped `machikoro.db.init.enabled` from `true` to `false` in `application.properties`. The stated intent — switching schema ownership to Flyway — is reasonable, but `DatabaseInitializer`'s `@ConditionalOnProperty` then disabled the whole bean, which also killed the only `Database.connect(dataSource)` call in the codebase. Exposed's static `TransactionManager` then has no default DB registered, so `transaction { ... }` throws on every DAO entry point.

## Why tests didn't catch it

`AbstractDBSetup` forces the property back to `\"true\"` via `@DynamicPropertySource`:

```kotlin
registry.add(\"machikoro.db.init.enabled\") { \"true\" }
```

So integration tests run with the bean enabled, masking the production default. (A follow-up should consider whether that override is more harm than help — without it, this would have failed loudly in CI the day 443b37f landed.)

## Fix

- Flip `machikoro.db.init.enabled` back to `true` in `application.properties` with a comment explaining why production needs it.
- Rewrite the KDoc on `DatabaseInitializer` so the next reader knows the bean's responsibility and what the property gates.
- Keep `SchemaUtils.create()` for now as a safety net — Flyway 11 + missing `flyway-database-postgresql` is a separate follow-up.

## Test plan

- [x] `./gradlew check` green locally with Docker up (409 tests pass, JaCoCo coverage gate passes).
- [x] Behavior matches pre-443b37f state. Property flip + comment-only changes.
- [x] After merge, manual smoke: launch server, tap Register in the Android client, confirm the database error no longer shows.

## Related

- Unblocks #254 (backend container restart recovery) — that issue requires a working end-to-end registration/login flow to demo.
- Follow-up issue to be filed: add `flyway-database-postgresql` so Flyway can actually own the schema and we can drop the `SchemaUtils.create()` safety net.

## Adjacent client-side smell (separate PR worth filing)

The Android Register dialog displays the raw server exception verbatim. That's both a UX issue (\"Database.connect()\" means nothing to users) and a minor information disclosure smell. Out of scope for this PR but worth a follow-up on the Client repo.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/database/DatabaseInit.kt
+++ b/src/main/kotlin/org/machikoro/server/database/DatabaseInit.kt
@@ -9,19 +9,24 @@ import org.springframework.stereotype.Component
 import javax.sql.DataSource
 
 /**
- * Initializes database schema when app starts
- * Automatically executed by SpringBoot because:
- * - Marked as a component
- * - Implements CommandLineRunner
+ * Registers the application's [DataSource] with Exposed's [Database] singleton
+ * so every `transaction { ... }` block in the DAOs has a default database to
+ * run against. Without this, the first DB-touching request throws
+ * "No database specified and no default database found".
+ *
+ * Also runs Exposed's [SchemaUtils.create] as a safety net for environments
+ * where Flyway hasn't (yet) executed — `CREATE TABLE IF NOT EXISTS` is
+ * idempotent so this is a no-op once Flyway has the schema covered. See
+ * tracker for the Flyway 11 + `flyway-database-postgresql` follow-up that
+ * makes Flyway the sole schema owner.
+ *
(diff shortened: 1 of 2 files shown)


eles17 (approved):
Approve

PR #262, opened 2026-05-14
Feat/256 flyway

### LINKED ISSUE #256: Add flyway-database-postgresql so Flyway can actually own the schema
## Symptom

Flyway is listed as a dependency (`org.flywaydb:flyway-core` 11.14.1, see `build.gradle.kts`) and the V1 baseline migration exists (`src/main/resources/db/migration/V1__init_schema.sql`), but Flyway has **never actually created the schema**. The reason the app works is Exposed's `SchemaUtils.create()` running inside `DatabaseInitializer` (see #PR-255). Pull the safety net out and the tables silently fail to materialize.

## Root cause

Flyway 10 and above split the database-specific dialects into separate modules. With only `flyway-core` on the classpath, Flyway can't resolve a `Database` implementation for Postgres at runtime — it logs a warning and quietly does nothing. The compose stack and Testcontainers both run Postgres, so we need `flyway-database-postgresql` alongside `flyway-core`.

Verified locally: removing `SchemaUtils.create(...)` from `initDatabase()` and re-running `./gradlew check` produces 154 failures with `relation \"player_cards\" does not exist` — proof that Flyway is not creating the tables on its own.

## Scope

1. Add `implementation(\"org.flywaydb:flyway-database-postgresql\")` to `build.gradle.kts`, pinned to the same `11.14.1` version (Spring Boot's BOM should manage this if we omit the version).
2. Update `gradle/verification-metadata.xml` with the new dependency + its transitive checksums (Gradle will fail until this is regenerated).
3. Confirm `./gradlew check` still passes — Flyway should now create the schema via V1 on a fresh Testcontainers Postgres.
4. **Drop `SchemaUtils.create()` from `DatabaseInit.initDatabase()`** so Flyway is the sole schema owner. `Database.connect(dataSource)` is the only line that stays.
5. Update `DatabaseInitTest` test names + bodies to reflect the new responsibility (no longer creates tables; just registers Exposed).
6. Consider removing the `machikoro.db.init.enabled` override from `AbstractDBSetup.registerProperties` so production-config drift fails CI loudly instead of being papered over by tests.

## Why this matters

- The `DatabaseInit.kt` KDoc currently has to apologise for the redundancy with a follow-up reference (\"See tracker for the Flyway 11 + flyway-database-postgresql follow-up\"). Closing this issue lets us delete that comment.
- Real Flyway support means future migrations (V2, V3, ...) actually get applied — not silently ignored.
- Once Flyway owns the schema, schema changes in PRs become explicit, reviewable migration files instead of edits to `Tables.kt` that depend on `SchemaUtils.create()` adding columns (which it can't — it only creates new tables, not new columns).

## Acceptance

- `./gradlew check` passes with `SchemaUtils.create()` removed from `initDatabase()`.
- Flyway logs at startup (`info` level) show V1 being applied on a fresh Testcontainers DB.
- `flyway_schema_history` table exists after first boot.
- The follow-up comment in `DatabaseInit.kt`'s KDoc is removed.

## Related

- PR #255 — production fix that left this gap exposed.
- #230 — original Flyway integration (added `flyway-core` but missed the dialect module).
- #443b37f — commit that first tried to switch ownership to Flyway and accidentally disabled `Database.connect()`.

Moritz2k1:
Implements Flyway Migrations and addresses the issue made by Fabian


FabianSolutions (approved):
LGTM. Merging

PR #284, opened 2026-05-18
Feat/278 seed cards

### LINKED ISSUE #278: Seed Card and Landmark Definitions at App Startup
Card and landmark definitions are missing from the database in production. 
TestDataSeeder seeds all 15 cards + 4 landmarks for tests only. 
Production has no equivalent — startGame() calls cardDao.findAll() and landmarkDao.findAll() expecting populated tables, causing silent failures or crashes on first game.

What to implement:
- Create a Spring ApplicationRunner or @PostConstruct bean (e.g. DataInitializer)
- Extract seeding logic from TestDataSeeder into this bean
- Seed all 15 establishments, their activation numbers, and all 4 landmarks using upsert/insertIgnore so it's idempotent on restart
- Bean runs once at app startup before any game can be started

Moritz2k1:
Implements the DataSeeder, which seeds cards and landmarks into the database when the game is initialized by the InitilizationService

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -129,6 +129,7 @@ class LobbyWebSocketController(
             payload = mapOf(
                 "playerId" to player.id,
                 "userId" to player.userId,
+                "username" to principal.username,
                 "gameId" to player.gameId,
                 "coins" to player.coins
             )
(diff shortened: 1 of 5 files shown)


eles17 (approved):
I checked the PR diff, GitHub reports it as mergeable, CI is green, SonarCloud passed, and local `./gradlew test` plus `./gradlew check` both pass. The startup ordering between `DatabaseInitializer` and `DataSeeder` looks correct, and the seeding is idempotent against the existing unique constraints.

One thing to track separately: now that all card definitions are seeded, gameplay logic for advanced card effects such as factories/market, TV Station, and Business Center should be handled in a follow-up if those effects are not already intentionally out of scope for this PR.

PR #288, opened 2026-05-18
fix: LobbyWebSocketController.kt fixed

### LINKED ISSUE #286: Lobby create/join events are broadcast to a global topic, auto-joining unrelated players into a lobby
## Summary

`LobbyWebSocketController.createLobby` and `joinLobby` both return their result with `@SendTo("/topic/public")` — a single global topic that every connected client subscribes to. Lobby events are therefore not scoped to a lobby: every logged-in player receives every other player's `LOBBY_CREATED` / `LOBBY_JOINED` events.

When a player creates a lobby, the `LOBBY_CREATED` broadcast reaches all other connected clients. The client reacts to it (`OkHttpWebSocketClient.handleLobbyCreated` sets `activeGameId`, after which `MainActivity` navigates to the Lobby screen), so a second, unrelated player is pulled into the creator's lobby without ever choosing to join. Because the two clients build their rosters from different events, the lobby views end up inconsistent between them.

## Steps to reproduce

1. Two players, both logged in, on the Home screen.
2. Player 1 creates a lobby.
3. Player 2 is moved into a lobby automatically, without entering a lobby code.
4. Player 1's lobby does not list Player 2; Player 2's lobby lists Player 1 (and shows them as ready).

## Expected vs. actual

- **Expected:** creating a lobby only affects the creator. Another player joins a lobby only by entering its code, and a join is visible to all members of that lobby — and only that lobby.
- **Actual:** lobby create/join events reach all connected clients via `/topic/public`; a second player is auto-joined into the creator's lobby, and the two players see different, inconsistent rosters.

## Where it happens

- `LobbyWebSocketController.createLobby` and `joinLobby` (`controller/LobbyWebSocketController.kt`) — both annotated `@SendTo("/topic/public")`.
- `/topic/public` is global; the client subscribes to it (`WebSocketContract.publicTopic`) and handles `LOBBY_CREATED` / `LOBBY_JOINED` from it for every lobby, not just its own.

## Suggested direction

Scope lobby events to a per-lobby topic instead of `/topic/public`:
- `LOBBY_CREATED` should go only to the creator (a user-specific reply), not a broadcast.
- `LOBBY_JOINED` should go to the specific lobby's topic (e.g. `/topic/game/{gameId}`, which the client already uses via `subscribeToGameTopic`), so only that lobby's members receive it.

The client must subscribe to and react to the per-lobby topic accordingly; a linked client-side issue may be needed for that coordination.

Moritz2k1:
Fixes the LobbyWebsocketController hopefully

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -11,13 +11,14 @@ import org.machikoro.server.service.LobbyService
 import org.slf4j.LoggerFactory
 import org.springframework.messaging.handler.annotation.MessageMapping
 import org.springframework.messaging.handler.annotation.Payload
-import org.springframework.messaging.handler.annotation.SendTo
 import org.springframework.messaging.simp.SimpMessageHeaderAccessor
+import org.springframework.messaging.simp.SimpMessagingTemplate
 import org.springframework.stereotype.Controller
 
 @Controller
 class LobbyWebSocketController(
     private val lobbyService: LobbyService,
+    private val messagingTemplate: SimpMessagingTemplate,
 ) {
     private val logger = LoggerFactory.getLogger(LobbyWebSocketController::class.java)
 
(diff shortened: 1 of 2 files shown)


FabianSolutions (approved):
LGTM. Merging

Issue #173, opened 2026-05-01
Determin Player order by Dice Logic


SoupCube:
### User Story
As a player, I want player order to be determined by dice rolls so that the starting order is fair and random.

### Acceptance Criteria
- Each player rolls a dice at the start of the game
- Dice results are random and between 1 and 6
- Results are shared with the players
- Player order is determined based on dice results
- In case of ties, a tie-breaker roll is triggered (if intended)

#### Related Issues
- https://github.com/SE2-Machi-Koro/Client/issues/98

Moritz2k1:
PlayerOrder gets determined by randomness. This issue is practically useless

eles17:
Keeping closed for now - if decided otherwise we reopen

PR #289, opened 2026-05-18
Fix/lobbyservice and controller

Moritz2k1:
Fixing LobbyService and Controller - needed for the issue #173 on the Client side


eles17 (approved):
Looks good overall. Approving 👍

Small non-blocking notes:

- getLobbyRoster could use a DTO long term instead of List<Map<String, Any>>, but it’s consistent with the current WebSocket payload style.
- Skipping players when UserDao.findById() returns null might be worth logging for debugging.
- Tests rely on exact send order, which is fine given the current implementation.

PR #290, opened 2026-05-19
287 joining a lobby does not deliver the existing roster to the joining player2.0

### LINKED ISSUE #287: Joining a lobby does not deliver the existing roster to the joining player
## Summary

When a player joins a lobby, `LobbyWebSocketController.joinLobby` emits a `LOBBY_JOINED` message that contains only the joining player. A player who joins an existing lobby is never sent the players who were already in it. There is no lobby-phase roster sync — on the client the full player list is only populated at `GAME_STARTED` or via the reconnect snapshot, both of which happen after the game has started. As a result, a joiner's lobby roster contains only themselves (plus anyone who joins later), so the creator and a joiner do not see a consistent player list.

## Relationship to #286

#286 fixes lobby events being broadcast to every connected client via the global `/topic/public`. It does not fix this defect: even with per-lobby topics, a joiner only subscribes to the lobby topic when they join, so they miss the join events of players already present. The joiner must be explicitly sent the current roster on join. (Today the global broadcast accidentally masks this in one direction — see #286 — which is why the auto-joined player currently does see the creator.)

## Steps to reproduce

1. Player 1 creates a lobby.
2. Player 2 joins the lobby by code.
3. Player 2's lobby roster does not include Player 1 — Player 2 was never sent the players already present.

## Expected vs. actual

- **Expected:** a player who joins a lobby sees every player currently in that lobby; the creator and a joiner see the same roster.
- **Actual:** a joiner only receives their own `LOBBY_JOINED`; players already in the lobby are never delivered to them.

## Where it happens

- `LobbyWebSocketController.joinLobby` (`controller/LobbyWebSocketController.kt`) — the `LOBBY_JOINED` payload carries only the joining player.
- `LobbyService` already has the data — `playerDao.getPlayers(gameId)` returns the full roster.

## Suggested direction

On a successful join, send the joining player the current lobby roster (a roster / lobby-state message listing all players currently in the lobby), in addition to announcing the new join to the lobby's topic. The client needs a handler to apply that roster during the lobby (waiting) phase, since today it only populates the roster at game start.

eles17:
## Summary
- Add a lobby roster DTO and DAO query that returns players with usernames in turn order.
- Send a private `LOBBY_ROSTER` snapshot to the joining player after successful `/app/lobby.join`.
- Keep the existing public `LOBBY_JOINED` broadcast unchanged for current lobby members.

## Changes
- Added `LobbyRosterDto` / `LobbyRosterPlayerDto`.
- Added `PlayerDao.getLobbyRoster(gameId)` joining `Players` with `Users`.
- Added `LobbyService.getLobbyRoster(gameId)` as the controller-facing roster API.
- Updated `LobbyWebSocketController.joinLobby` to send:
  - private `/queue/lobby-user{sessionId}` `LOBBY_ROSTER`
  - public `/topic/game/{gameId}` `LOBBY_JOINED`
- Added/updated tests for controller behavior, service delegation, and DAO roster ordering.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -3,6 +3,7 @@ package org.machikoro.server.controller
 import io.github.springwolf.core.asyncapi.annotations.AsyncListener
 import io.github.springwolf.core.asyncapi.annotations.AsyncOperation
 import org.machikoro.server.auth.userPrincipal
+import org.machikoro.server.dto.LobbyRosterDto
 import org.machikoro.server.dto.MessageType
 import org.machikoro.server.dto.WebSocketMessage
 import org.machikoro.server.exception.CustomWebSocketException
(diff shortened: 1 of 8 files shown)


Moritz2k1 (approved):
Looks good. Merged

PR #312, opened 2026-05-21
fix(session): clarify reconnect state and dice roll completion

### LINKED ISSUE #291: fix: validate lobby/game transition and dice roll state responses
## Problem
Manual testing showed the client unexpectedly skipping screens and the dice roll never settling. The visible symptoms are in the client, but the server should verify that session, lobby, game, and dice-roll state transitions are consistent and unambiguous.

## Observed client behavior
1. The app moved from the start screen to the lobby screen without user interaction.
2. After returning to the start screen and logging in, the app skipped the main/lobby-creation screen and went directly into the game screen.
3. Clicking `Würfeln` caused the dice to continuously spin/change numbers instead of stopping.

## Expected backend behavior
- A user/session should not be reported as already in a lobby or game unless that state is valid and current.
- Lobby and game transition endpoints/events should make it clear whether the user should be on start, main/lobby creation, lobby, or game screen.
- Dice roll handling should produce one completed result per valid roll request and expose a stable completed state to the client.
- Invalid duplicate roll requests should be rejected or ignored predictably.

## Suggested investigation
- Check whether stale sessions, stale lobby memberships, or unfinished games are returned after login/reconnect.
- Verify lobby-to-game transition rules and whether a player can be treated as already in-game after leaving/back navigation.
- Inspect dice roll endpoint/event behavior for repeated emissions, missing completion state, or duplicate roll handling.
- Add or update tests for login/reconnect state, lobby membership cleanup, game start transitions, and dice roll completion.

valentinaschiavon99:
## Summary
- Add server-resolved session state to login responses so clients can distinguish main, lobby, and game screens from valid current memberships.
- Resolve reconnect membership by preferring in-progress games, falling back to waiting lobbies, and ignoring finished games.
- Serialize dice rolls per game, reject duplicate completed rolls predictably, and broadcast stable completed roll payloads.
- Derive the rolling player from the authenticated active turn instead of trusting client-supplied player IDs.
- Accept the current client legacy wrapped dice payload, payload.diceCount, for backward-compatible roll requests.
- Unblocks https://github.com/SE2-Machi-Koro/Client/issues/175

## Tests
- ./gradlew test --tests org.machikoro.server.service.DiceServiceTests --tests org.machikoro.server.controller.GameControllerTest --tests org.machikoro.server.controller.AuthControllerTests --tests org.machikoro.server.service.GameSyncServiceTest --tests org.machikoro.server.service.GameStateGuardTest --tests org.machikoro.server.dto.RollDiceRequestTests

## Client compatibility
- The current Client checkout ignores the new login sessionState field safely because its auth JSON parser uses ignoreUnknownKeys = true.
- The dice endpoint remains compatible with the current Client payload shape while keeping server-side player ownership authoritative.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/AuthController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/AuthController.kt
@@ -7,6 +7,7 @@ import org.machikoro.server.dto.LogoutRequest
 import org.machikoro.server.dto.RegisterRequest
 import org.machikoro.server.exception.InvalidCredentialsException
 import org.machikoro.server.service.AuthService
+import org.machikoro.server.service.GameSyncService
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
@@ -18,7 +19,10 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 @RequestMapping("/auth")
 @Tag(name = "Authentication", description = "User registration, login, and session management")
-class AuthController(private val authService: AuthService) {
+class AuthController(
+    private val authService: AuthService,
+    private val gameSyncService: GameSyncService,
+) {
     private val logger = LoggerFactory.getLogger(AuthController::class.java)
 
     @PostMapping("/register")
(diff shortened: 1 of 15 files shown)


eles17:
@valentinaschiavon99 did you test the running game?
This PR changes live client/server behavior: login now returns sessionState, dice rolling now ignores client-supplied playerId, accepts legacy wrapped dice payloads, changes duplicate-roll behavior, and broadcasts richer roll payloads. Those are exactly the kinds of server-only changes that can still break the client at runtime.

valentinaschiavon99:
> @valentinaschiavon99 did you test the running game?
> This PR changes live client/server behavior: login now returns sessionState, dice rolling now ignores client-supplied playerId, accepts legacy wrapped dice payloads, changes duplicate-roll behavior, and broadcasts richer roll payloads. Those are exactly the kinds of server-only changes that can still break the client at runtime.
 
I intentionally kept the server backward-compatible with the current Client instead of requiring a coordinated Client refactor.
The server now derives the rolling player from the authenticated active turn, so it no longer trusts client-supplied playerId. At the same time, it still accepts the Client’s existing wrapped dice payload shape, including payload.diceCount, to avoid breaking the current app while making the server-side ownership and duplicate-roll handling authoritative.

FabianSolutions (approved):
LGTM. Merging

PR #315, opened 2026-05-22
refactor(auth): move DTO validation to JSR-303 annotations

### LINKED ISSUE #170: Move auth DTO validation from require() to JSR-303 annotations

FabianSolutions:
## Summary
Closes #170 (follow-up from the PR #167 review). Moves auth input validation from imperative `require(...)` calls in `AuthService` to declarative JSR-303 constraints on the request DTOs, validated at the controller boundary via `@Valid`.

- Adds `spring-boot-starter-validation` (`gradle.lockfile` + `gradle/verification-metadata.xml` regenerated for the new hibernate-validator / jakarta artifacts).
- `@field:NotBlank` / `@field:Size` on `RegisterRequest` / `LoginRequest` / `LogoutRequest`, with messages matching the previous `require(...)` text verbatim.
- `@Valid` on the three `AuthController` `@RequestBody` params.
- New `ValidationExceptionHandler` (`@RestControllerAdvice`) flattens `MethodArgumentNotValidException` to the existing **plain-text 400** body — option (b) from the issue — so the Android client contract (Client #88/#92) is unchanged.
- Removes the now-redundant `require(...)` from `AuthService`; the `trim().lowercase()` username normalisation stays.

## Notes
- `@Size(max=50/72)` validates raw string length, matching the old char-based checks (the username check was previously on the trimmed value — a negligible difference only for whitespace-padded 50+ char usernames).
- The advice returns the first field error's message; for single-field violations (the realistic case) this is deterministic. Multi-field ordering is the contract nuance the issue acknowledged.
- No conflict with `GlobalWebSocketExceptionHandler` (STOMP `@MessageExceptionHandler`, separate pipeline).

## Test plan
- [x] `./gradlew build` green locally **with Docker** — full suite incl. Testcontainers integration tests + `jacocoTestCoverageVerification`.
- [x] `AuthServiceTest` — removed the 7 `require()`-based cases (validation no longer lives in the service); normalization / duplicate / credential cases retained.
- [x] `AuthControllerTests` — removed the 2 service-mocked validation cases.
- [x] New `AuthValidationTest` (MockMvc, full context) — asserts `400` + exact plain-text body for blank username/password/sessionToken and over-length username/password.
- [ ] Manual (optional): `POST /auth/register` `{"username":"","password":"x"}` → `400` body `Username must not be blank` (plain text, not JSON).


Moritz2k1 (approved):
Loosk fine to me. Implements the JSR 303 norm

PR #333, opened 2026-05-26
331 fix broadcast lobby roster after filllobby so host sees dummy players

kutschera13:
## Problem
When the host pressed [Debug] Fill with dummies, the dummy players were 
added successfully but the host's lobby screen stayed empty because no 
LOBBY_ROSTER update was sent after the fill.

## Root Cause
DebugService.fillLobby() only broadcast LOBBY_JOINED per dummy player. 
The host never receives these as a roster update.

## Fix
After all dummies are added, broadcast a full LOBBY_ROSTER message to 
/topic/game/{gameId} so all connected clients (including the host) 
receive the complete updated player list.

## Changes
- `src/main/kotlin/org/machikoro/server/service/DebugService.kt`
- `src/test/kotlin/org/machikoro/server/service/DebugServiceTest.kt`

## Testing
All tests pass ✅

## Note
⚠️ After merging, a new Docker image needs to be built and deployed.
Please merge as soon as possible so the CI/CD pipeline can build the 
new image and the fix can be tested on the server.

After merge, run on the server:
docker compose pull
docker compose up -d

### DIFF
--- a/src/main/kotlin/org/machikoro/server/service/DebugService.kt
+++ b/src/main/kotlin/org/machikoro/server/service/DebugService.kt
@@ -2,6 +2,7 @@ package org.machikoro.server.service
 
 import org.machikoro.server.dao.UserDao
 import org.machikoro.server.dto.DebugSeedResponse
+import org.machikoro.server.dto.LobbyRosterDto
 import org.machikoro.server.dto.LoginResponse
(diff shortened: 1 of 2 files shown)


Moritz2k1:
image

I have opened up a PR on the Client side that fixes the problems without needing to adjust the backend

PR #335, opened 2026-05-30
Feat/334 add admin accounts

### LINKED ISSUE #334: Add isAdmin Flag into Users Table
### Goal

Add isAdmin Entry into the Users Table to be able to purge the database on the Uni Server / Protected environments.
Seed 4 Admin Accounts into the database.
Admin Accounts should only work if the debug_flag is set to true in the .env file

Moritz2k1:
Added an isAdmin column to the UsersTable
Database gets seeded with 4 admin accounts now

Now only admin accounts can use the debug endpoints, since we currently get a 403 when using a debug endpoint on the uni server

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,3 +22,4 @@ PGADMIN_PASSWORD=admin
 
 #Debugging
 DEBUG_ENABLED=true
+ADMIN_PASSWORD=adminacc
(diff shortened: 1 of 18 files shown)


FabianSolutions (approved):
LGTM. Fixing uni server returning 403 on debug endpoints because nothing knew "admin"

PR #191, opened 2026-05-02
180 round counter logic

doniliks:
## Summary
Adds roundsPlayed to EndTurnOutcome.Won from game state
Sent to clients in broadcastWin (renamed from broadcastWinner since will be expanded in future with e.g ranking, time etc.
Enables displaying total rounds on the winner screen
Existing tests were adjusted respectivelly 

#180 was already partly finished, just small feature to complete

Closes #180

PR #337, opened 2026-05-31
feat: debug-only end-game endpoint (#305)

### LINKED ISSUE #305: OPTIONAL: debug endpoint to end the game with the caller as winner
## Task

Add a debug-only server endpoint that immediately ends the current game with the calling player as the winner, so testers can reach the winner screen without playing a game through. Reuses the existing `GAME_END` broadcast — no new event types needed on the wire; the client's existing winner-screen path (PR #205) handles the response unchanged.

## Scope

- Add a new endpoint on `DebugController` (e.g. `POST /debug/end-game` with `{ gameId: Int }`).
- Gating (three layers, all already in place):
  - `@ConditionalOnProperty(debug.enabled=true)` (inherited from `DebugController`) — endpoint is not registered when `debug.enabled=false`.
  - `debugService.validateAdmin(authHeader)` — caller must hold an admin session token (same pattern added in #335).
- Validation:
  - The calling user is a player in the requested game.
  - The game is `IN_PROGRESS` (not `WAITING`, not `FINISHED`).
- Mutation: transition the game to `FINISHED` with the calling player as winner via the existing end-game logic in `GamePhaseService` (don't duplicate the path — reuse the production code so the result is indistinguishable from a natural win).
- Broadcast: emit the existing `GAME_END` event on `/topic/game/{gameId}` carrying the winner — no new typed event needed.

## Acceptance Criteria

- Given `debug.enabled=true` and an admin caller in an `IN_PROGRESS` game, when the endpoint is called, then the game becomes `FINISHED` with the caller as winner and every subscriber to `/topic/game/{gameId}` receives a `GAME_END` broadcast.
- Given `debug.enabled=false`, the endpoint is not registered (request returns 404).
- Given the caller is not an admin, the endpoint returns 403 and the game state is unchanged.
- Given the caller has no or invalid session token, the endpoint returns 401 and the game state is unchanged.
- Given the caller is not a player in the requested game, the endpoint returns a structured 4xx error and the game state is unchanged.
- Given the game is already `FINISHED` or still `WAITING`, the endpoint returns a structured 4xx error and the game state is unchanged.
- Logged at `INFO` on success, `WARN` on rejected attempts (with reason).
- No raw database entities returned over the wire.
- Sealed Kotlin types are used for any new request/error models.

## Definition of Done

- Endpoint added to `DebugController`, reusing the existing admin-auth helper.
- End-game transition delegates to existing `GamePhaseService` win-handling — does not duplicate winner logic.
- Unit tests cover: success (state + broadcast), 401, 403, 404 (when debug disabled), not-in-game, game-not-in-progress, game-finished.
- Integration test covers the WebSocket broadcast end-to-end (mirror the existing `GameWebSocketBroadcastIntegrationTest` style).
- No `println()`, no broad `catch (Exception)`, no raw nulls returned.
- Existing Gradle tests + SonarCloud + CodeQL pass.

## Notes

- For testing only. In a multiplayer game, calling this ends the game for everyone immediately with the caller as winner — opponents lose without warning. The admin-only gating keeps this off-limits to regular users.
- Pairs with Client #191 (the UI button that calls this endpoint).

FabianSolutions:
## What

Adds a debug-only `POST /debug/end-game` endpoint that force-ends a running game with the admin caller as the winner, so testers can reach the winner screen without playing a full game. Closes #305.

The client-side debug button that calls this is tracked separately in Client #191.

## How it's gated (layered)

- `@ConditionalOnProperty(debug.enabled=true)` — the whole `DebugController` is unregistered in production.
- `validateAdmin(authHeader)` — requires a valid **admin** session token (401 / 403 otherwise).
- The caller must be a **player in the game** — otherwise `NotInGameException` → 422.

## Design

Two commits:

1. **`refactor: extract GameEndBroadcaster`** — the GAME_END broadcast was inlined in `GameController.broadcastWin`. Extracted to a `GameEndBroadcaster` service so the production end-of-game path and this new endpoint emit an identical wire shape (same `MessageType`, payload keys, and `/topic/game/{id}`). No behaviour change on the production path.
2. **`feat: add POST /debug/end-game`** — the endpoint, service method, request DTO, and exception.

`DebugService.endGame` is `@Transactional` and applies the **same win side effects as `GamePhaseService.endTurn`** (increment wins, increment games-played for all, status → `FINISHED`) but bypasses `WinConditionService.detectWinner` so no landmarks are required. The controller then broadcasts GAME_END and runs `cleanupFinishedGameData` **post-commit**, mirroring `GameController.endTurn` exactly. `validateAdmin` now returns the resolved `UserModel` so the controller threads the admin into `endGame` with a single token lookup.

Status codes: `200` (Won), `401` (no/invalid token), `403` (not admin), `404` (game not found), `422` (not a player / game not running), `500` (unexpected).

## Testing

- `./gradlew check` — full unit + Testcontainers integration suite + 80%-per-class coverage gate: **green**. New/changed classes are at 100% line coverage.
- **Manual 2-emulator e2e** against a local backend: two admin clients in a live game, triggered the endpoint; **both clients navigated to the winner screen** showing the caller as winner, the game + player rows were cleaned up, stats updated correctly, and there were **no client-side errors** despite the game being deleted (the client renders the winner screen from the GAME_END snapshot in the payload).

---

### Note: `debug.enabled=false` returns 403, not 404 (documented deviation)

#305's AC says the disabled endpoint should return `404`. In practice it returns **`403`**, which is a *stronger* "not accessible" guarantee, for two compounding reasons:

1. `@ConditionalOnProperty(debug.enabled=true)` means the `DebugController` bean is never registered when debug is off — so there is no handler.
2. `SecurityConfig` deliberately only adds `/debug/**` to `permitAll` + CSRF-exempt when `debug.enabled=true`. With debug off, a `POST /debug/end-game` falls through to `anyRequest().authenticated()` + CSRF and is rejected with `403` **before** the dispatcher could ever produce a `404`.

We chose to keep the `403` (verified by `DebugEndGameDisabledIntegrationTest`) rather than loosen `SecurityConfig` to `permitAll` `/debug/**` unconditionally just to surface a `404` — that would remove an intentional defense-in-depth layer. The endpoint is verifiably unreachable when debug is disabled either way.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/DebugController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/DebugController.kt
@@ -138,8 +138,9 @@ class DebugController(
                 .body(DebugEndGameError("NOT_ADMIN", e.message ?: "Admin access required"))
         } catch (e: GameNotFoundException) {
             logger.warn("Debug end-game rejected: {}", e.message)
+            // GameNotFoundException extends CustomWebSocketException, whose message is non-null.
             ResponseEntity.status(HttpStatus.NOT_FOUND)
-                .body(DebugEndGameError("GAME_NOT_FOUND", e.message ?: "Game not found"))
+                .body(DebugEndGameError("GAME_NOT_FOUND", e.message))
         } catch (e: NotInGameException) {
             logger.warn("Debug end-game rejected: {}", e.message)
             ResponseEntity.status(HttpStatus.UNPROCESSABLE_ENTITY)
--- a/src/main/kotlin/org/machikoro/server/controller/DebugController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/DebugController.kt
@@ -2,9 +2,9 @@ package org.machikoro.server.controller
 
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.tags.Tag
+import org.machikoro.server.dto.DebugEndGameError
 import org.machikoro.server.dto.DebugSeedResponse
 import org.machikoro.server.dto.EndGameRequest
-import org.machikoro.server.dto.EndTurnOutcome
 import org.machikoro.server.dto.FillLobbyRequest
 import org.machikoro.server.dto.LoginResponse
 import org.machikoro.server.exception.CustomWebSocketException
@@ -121,7 +121,7 @@ class DebugController(
     fun endGame(
         @RequestHeader("Authorization", required = false) authHeader: String?,
         @RequestBody request: EndGameRequest,
-    ): ResponseEntity<EndTurnOutcome.Won> {
+    ): ResponseEntity<Any> {
         return try {
             val caller = debugService.validateAdmin(authHeader)
             val outcome = debugService.endGame(request.gameId, caller)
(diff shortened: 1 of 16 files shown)


eles17 (commented):
Thanks, this looks functionally aligned with #305 overall: the branch adds `POST /debug/end-game`, keeps it debug/admin gated, validates the caller is in a running game, emits the existing `GAME_END` payload, and the focused unit tests pass.

I would like a few DoD gaps clarified or fixed before merge:

- The issue asked to reuse the production end-game/win-handling path. `DebugService.endGame` currently duplicates the win side effects (`incrementWins`, `incrementGamesPlayed`, `updateStatus`) instead of delegating to shared `GamePhaseService` logic. It works now, but could drift from the real win path later.
- Rejected cases currently return status-only `401/403/404/422` responses. #305 asks for structured 4xx errors.
- The new endpoint has a broad `catch (Exception)`, while the DoD says no broad catch.
- I don’t see the requested WebSocket integration test or explicit `debug.enabled=false` 404 coverage.

CI is green and the main happy path seems sound, but I’d prefer addressing these acceptance-criteria gaps before merging, or explicitly documenting that we accept them as out of scope.

eles17 (commented):
Thanks for addressing the review comments. The main gaps are fixed now:

- `DebugService.endGame` now delegates win side effects through `GamePhaseService.finishGameWithWinner(...)`.
- `POST /debug/end-game` now returns structured 4xx bodies via `DebugEndGameError`.
- The broad `catch (Exception)` was removed from the new endpoint.
- WebSocket coverage was added with `DebugEndGameBroadcastIntegrationTest`.

One thing I’d still like clarified before approval: #305 says `debug.enabled=false` should make the endpoint return `404`, but `DebugEndGameDisabledIntegrationTest` now asserts `403` because security rejects `/debug/**` before dispatcher routing. If we accept `403` as the intended behavior, please update the issue/PR notes to explicitly document that deviation. Otherwise, adjust security/controller routing so the disabled endpoint produces the expected `404`.

Also, SonarCloud shows “1 New issue” even though the quality gate passed. Please check whether that issue is relevant or false-positive before merge.

FabianSolutions:
Thanks — both points addressed:

**SonarCloud "1 New issue" (S6619):** it was a real dead-code smell, not a false positive. The `GameNotFoundException` branch did `e.message ?: "Game not found"`, but `GameNotFoundException` extends `CustomWebSocketException` (non-null `message`), so the fallback could never fire. Removed it (595d23d). The other catch branches keep their elvis because they catch `RuntimeException` subclasses with a nullable `message`.

**403 vs 404 on `debug.enabled=false`:** documented as an accepted deviation (added a "Note" section to the PR description). Short version: the `403` is a *stronger* guarantee than `404` — `@ConditionalOnProperty` means no controller bean exists, **and** `SecurityConfig` only opens `/debug/**` when debug is enabled, so security rejects the path before routing. Producing a literal `404` would mean `permitAll` + CSRF-exempting `/debug/**` unconditionally, which removes an intentional defense-in-depth layer, so we kept the `403` (asserted by `DebugEndGameDisabledIntegrationTest`). Happy to switch to the `404` approach if you'd prefer strict AC compliance over the stricter security posture.

eles17 (approved):
Thanks, that addresses my remaining concerns.

I’m okay with the documented `403` deviation for `debug.enabled=false`; keeping `/debug/**` closed at the security layer is the better tradeoff here.

Approved from my side.

PR #342, opened 2026-06-03
feat(DiceService): validate diceCount range and unify two-dice resolu…

### LINKED ISSUE #307: feat: accept diceCount parameter in roll dice request
## Description
The `RollDiceRequest` DTO currently does not support a `diceCount` field.
To support the train station landmark mechanic, the server must accept
how many dice the player wants to roll.

## Acceptance Criteria
- `RollDiceRequest` includes a `diceCount` field (integer)
- Valid values are `1` or `2`
- If `diceCount` is not provided, default to `1`
- If `diceCount` is outside the valid range, reject with a descriptive error

### LINKED ISSUE #308: feat: validate train station ownership before allowing 2 dice
## Description
A player may only roll 2 dice if they have built the train station landmark.
The server must enforce this rule to prevent clients from bypassing the
restriction by sending `diceCount = 2` without having built the train station.

## Acceptance Criteria
- Before processing a roll with `diceCount = 2`, check that the requesting
  player has `TRAIN_STATION` landmark with `isBuilt = true`
- If the player does not own the built train station and sends `diceCount = 2`,
  reject the request with an appropriate error response
- If `diceCount = 1`, no landmark check is required

## Notes
- Train station state is stored in `PlayerLandmarkDao`
- The check must use the requesting player's `userId` from the authenticated session

### LINKED ISSUE #309: feat: roll multiple dice and calculate sum for earnings resolution
## Description
When `diceCount = 2`, the server must roll two dice independently and
use their sum to determine which establishments pay out.
Per the official rules: "bei 2 Würfeln gilt nur die Summe der beiden
Würfel, nicht die einzelnen Würfelzahlen."

## Acceptance Criteria
- When `diceCount = 1`, roll one die (result between 1–6)
- When `diceCount = 2`, roll two dice independently (each between 1–6)
- The **sum** of both dice is used for earnings resolution
- The `DiceService` must support rolling a variable number of dice
- Existing single-die roll logic must not be broken

## Notes
- The Freizeitpark (amusement park) landmark gives the active player
  an extra turn when rolling doubles (Pasch) with 2 dice — this is
  out of scope for this issue but the dice result list must support it
  in future

### LINKED ISSUE #310: feat: include individual dice results in roll dice response
## Description
The `RollDiceResponse` must return the individual dice results as a list
so the client can display each die separately and animate them correctly.
The sum is used for game logic; the individual values are used for UI display.

## Acceptance Criteria
- `RollDiceResponse` includes a `result` field of type `List<Int>`
  containing the individual dice values (e.g. `[3, 4]` for two dice,
  `[5]` for one die)
- The existing `lastDiceRoll` field in the game state stores the **sum**
  of all dice for reconnect snapshots
- The WebSocket broadcast includes the full `result` list so the client
  can render both dice

## Notes
- The client already parses `payload.result` as a `List<Int>` in
  `parseDiceResult()` in `OkHttpWebSocketClient.kt`
- Single-die rolls return a list with one element: `[n]`

kutschera13:
## Summary
Extends the roll-dice flow to support the Train Station landmark mechanic. Players with a built Train Station can now choose to roll 1 or 2 dice. The server validates the request, enforces the landmark rule, and broadcasts the individual dice results to all clients.

## Changes
- `DiceService.kt`: Added `diceCount` range validation (must be 1 or 2), unified two-dice resolution logic
- `RollDiceRequest.kt`: `diceCount` field already present — now used as primary input instead of legacy fallback

## Acceptance Criteria
- ✅ `RollDiceRequest` accepts `diceCount` (integer, default 1)
- ✅ Valid values are 1 or 2 — invalid values rejected with `INVALID_DICE_COUNT` error
- ✅ `diceCount = 2` requires built Train Station — rejected with `NO_TRAIN_STATION` if not built
- ✅ `diceCount = 1` requires no landmark check
- ✅ Two dice rolled independently, sum used for earnings resolution
- ✅ WebSocket broadcast includes `result` as `List<Int>` for client UI rendering

## Related Issues
Closes #307
Closes #308
Closes #309
Closes #310


Moritz2k1:
Sonarcloud says to refactor a method to reduce the cognitive complexity. Can you fix this - then I will merge it

eles17:
@kutschera13 i see you really mastered Github 🥇

eles17 (changes requested):
Thanks Moritz for reviewing. Besides the Sonar cognitive complexity refactor, I don’t think all linked sub-issues are fully addressed yet.

Could you please also add/fix:

- Make `diceCount` the primary source of truth. If `diceCount = 1` is explicitly sent, it should roll one die even if legacy `rollTwoDice = true` is also present.
- Add/adjust tests for invalid `diceCount` values, e.g. `0`, `3`, `-1`, expecting `INVALID_DICE_COUNT`.
- Add a test that `diceCount = 2` without a built Train Station is rejected with `NO_TRAIN_STATION`.
- Add a test that `diceCount = 1` does not require/check Train Station.
- Verify issue #310’s DTO contract: it asks for `RollDiceResponse.result: List<Int>`, but the DTO currently exposes `dice: List<Int>`. The WebSocket payload sends `result`, but the response model itself does not match the acceptance criteria.

So I think #307-#310 should not be considered fully closed until those points are covered.

eles17 (approved):
Thanks for covering the requested points. Approved.

PR #344, opened 2026-06-03
Sync Player Ready State in Lobby

### LINKED ISSUE #343: Implement player ready state for lobby
### Problem
Backend has no concept of player ready state.
Clients can press ready, but no server endpoint exists - the toggle is silently dropped

Frontend sends isReady toggle to /app/lobby.ready, which does not exist on the server

### Goal
Store ready state in memoy:

LobbyService                                                                                                                                                                                                                                                                                                     
- Add ConcurrentHashMap<Int, Set<Int>> (readyPlayers: gameId → set of ready userIds)                                                                                                                                                                                                                             
- Add setReadyState(gameId, userId, isReady): List<LobbyRosterPlayerDto>                                                                                                                                                                                                                                         
- Update startGame() to throw NotAllPlayersReadyException if any player not in ready set                                                                                                                                                                                                                         
- Clear ready set in cleanUpLobby() and reset all when a new player joins                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                  LobbyRosterPlayerDto                                                                                                                                                                                                                                                                                             
- Add isReady: Boolean (derived from in-memory set when building the roster)                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                   
LobbyWebSocketController                                                                                                                                                                                                                                                                                         
- Add @MessageMapping("/lobby.ready") handler: reads gameId + isReady from payload, calls setReadyState(), broadcasts LOBBY_ROSTER to /topic/game/{gameId}                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
NotAllPlayersReadyException                                                                                                                                                                                                                                                                                      
- New exception, errorCode = "NOT_ALL_PLAYERS_READY"

Moritz2k1:
Problem:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 Two bugs in the lobby ready flow:                                                                                                                                                                                                                                                                                
- Toggling ready on one client had no effect on other clients — the backend had no /app/lobby.ready endpoint, so the toggle was silently dropped                                                                                                                                                                 
- The host could start the game without all players being ready — startEnabled only checked if the host was ready, not everyone                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                   
Changes:
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
- New /app/lobby.ready WebSocket endpoint — updates in-memory ready state, broadcasts LOBBY_ROSTER with isReady flags to all lobby members                                                                                                                                                                       
- LobbyRosterPlayerDto — added isReady: Boolean                                                                                                                                                                                                                                                                  
- LobbyService — in-memory readyPlayers map (gameId → set of ready userIds), setReadyState(), getLobbyRoster() enriched with ready flags, startGame() rejects with NOT_ALL_PLAYERS_READY if any player not ready                                                                                                 
- LobbyWebSocketController — broadcasts full LOBBY_ROSTER after LOBBY_JOINED and LOBBY_LEFT so ready states stay visible when lobby composition changes                                                                                                                                                          
- leaveLobby() — returns LobbyDeleted instead of throwing PlayerNotFoundException when player/game already gone (fixes crash after debug purge)

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -5,7 +5,6 @@ import io.github.springwolf.core.asyncapi.annotations.AsyncOperation
 import org.machikoro.server.auth.userPrincipal
 import org.machikoro.server.dto.LobbyLeavingOutcome
 import org.machikoro.server.dto.LobbyRosterDto
-import org.machikoro.server.dto.LobbyRosterPlayerDto
 import org.machikoro.server.dto.MessageType
 import org.machikoro.server.dto.WebSocketMessage
 import org.machikoro.server.exception.CustomWebSocketException
@@ -162,7 +161,8 @@ class LobbyWebSocketController(
             )
         }
 
-        // Broadcast join to all members already subscribed to this lobby's topic
+        // Broadcast join event to existing members, then follow up with full roster so
+        // ready states remain accurate for everyone in the lobby
         messagingTemplate.convertAndSend(
             "/topic/game/${player.gameId}",
             WebSocketMessage(
@@ -179,28 +179,19 @@ class LobbyWebSocketController(
                 )
             )
         )
+        val updatedRoster = lobbyService.getLobbyRoster(player.gameId)
+        messagingTemplate.convertAndSend(
+            "/topic/game/${player.gameId}",
+            WebSocketMessage(
+                type = MessageType.LOBBY_ROSTER,
+                sender = "SERVER",
+                content = "Lobby roster",
+                gameId = player.gameId,
+                payload = LobbyRosterDto(players = updatedRoster)
+            )
+        )
     }
 
-    /**
-     * Handles a player leaving a lobby via WebSocket.
-     *
-     * Client sends a message to `/app/lobby.leave` with the `gameId`
-     * in the payload.
-     *
-     * The authenticated player is removed from the lobby. If the lobby
-     * still contains real players, a `LOBBY_LEFT` event is broadcast to
-     * `/topic/game/{gameId}` so remaining members can update screen.
-     *
-     * If no real players remain (debug/dummy players do not count), the
-     * lobby is deleted and a `HOST_LEFT` event is broadcast so clients
-     * can close the lobby UI.
-     *
-     * Authentication is derived from the WebSocket principal; the sender
-     * field in [WebSocketMessage] is ignored to prevent spoofing.
-     *
-     * @throws CustomWebSocketException when authentication is missing,
-     * the payload is invalid, or `gameId` is absent.
-     */
     /**
      * Handles a player toggling their ready state in the lobby.
      *
@@ -254,6 +245,26 @@ class LobbyWebSocketController(
         )
     }
 
+    /**
+     * Handles a player leaving a lobby via WebSocket.
+     *
+     * Client sends a message to `/app/lobby.leave` with the `gameId`
+     * in the payload.
+     *
+     * The authenticated player is removed from the lobby. If the lobby
+     * still contains real players, a `LOBBY_LEFT` event is broadcast to
+     * `/topic/game/{gameId}` so remaining members can update screen.
+     *
+     * If no real players remain (debug/dummy players do not count), the
+     * lobby is deleted and a `HOST_LEFT` event is broadcast so clients
+     * can close the lobby UI.
+     *
+     * Authentication is derived from the WebSocket principal; the sender
+     * field in [WebSocketMessage] is ignored to prevent spoofing.
+     *
+     * @throws CustomWebSocketException when authentication is missing,
+     * the payload is invalid, or `gameId` is absent.
+     */
     @MessageMapping("/lobby.leave")
     @Suppress("UNUSED_PARAMETER")
     fun leaveLobby(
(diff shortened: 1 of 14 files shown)


FabianSolutions (changes requested):
## Review: requesting changes

Thanks for this — the happy-path feature is well-built: clean wire contract, thread-safe `ConcurrentHashMap.newKeySet` usage, accurate `toggleReady` KDoc, and precise broadcast-count/ordering tests. However, there are a few blocker-class issues that should be resolved before merging. Note that CI stays green here largely because the debug tests mock `startGame`, so the new gate is never actually exercised.

### Blockers

**1. `leaveLobby` can falsely broadcast `HOST_LEFT` to a live lobby.**
`leaveLobby` now returns `LobbyLeavingOutcome.LobbyDeleted(gameId)` when the player or game isn't found, **without calling `cleanUpLobby()`** — so the lobby is still alive (`LobbyService.kt:206-210`). The controller maps every `LobbyDeleted` to a `HOST_LEFT` broadcast on the shared `/topic/game/{gameId}` with no membership check (`LobbyWebSocketController.kt:294,326-342`), and `StompAuthChannelInterceptor` only guards `CONNECT` (SEND/SUBSCRIBE to any game topic pass through). Consequences:
- Any authenticated user can send `/app/lobby.leave` with another lobby's (sequential, enumerable) `gameId` and force all real members to think the host left.
- Even without an attacker, a duplicate/stale leave frame (client retry after already leaving) re-broadcasts `HOST_LEFT` to remaining members.

On `main` these paths threw `CustomWebSocketException` routed to `@SendToUser("/queue/errors", broadcast=false)` (caller-only, no topic broadcast), so this is a new regression.
*Suggested fix:* only return `LobbyDeleted` when `cleanUpLobby` actually ran; for the not-found / not-a-member case return a distinct caller-only no-op outcome that does not broadcast to the topic.

**2. The new "all players ready" `startGame` gate breaks debug tooling and has no host auto-ready.**
The gate requires every row in `playerDao.getPlayers(gameId)` to be in the in-memory `readyPlayers` set (`LobbyService.kt:303-316`), but the only writer is the human `/app/lobby.ready` endpoint. So:
- `DebugService.seed()` adds 4 debug players then calls the real `startGame` -> `NOT_ALL_PLAYERS_READY` -> `/debug/seed` returns HTTP 500. Same for `/debug/fill-lobby`. Deterministic, no workaround (bots have no client to ready up). Uncaught because `DebugServiceTest` mocks `lobbyService.startGame`.
- The host is never auto-readied, so even a normal all-human game starts only if the client explicitly sends `lobby.ready` for every player including the host. The PR's own integration tests had to add `setReadyState(host, true)` calls that weren't previously needed.

*Suggested fix:* exclude `DEBUG_USER_PREFIX` players from the gate (use `getLobbyRoster`, which carries usernames — mirrors the existing `noRealPlayersRemain` filter) and/or auto-ready debug players; confirm or auto-ready the host. Add a test that starts a debug-filled lobby.

### Majors (a single guard resolves most of these)

- **No membership/existence/phase validation in `toggleReady`/`setReadyState`** (`LobbyService.kt:117-121`). Any authenticated user can `setReadyState` for any `gameId`, triggering unsolicited `LOBBY_ROSTER` broadcasts to any game topic. Inconsistent with `addUserToLobby` and `GameStateGuard`. (The start-gate itself is sound against spoofed userIds — it is anchored to DB players — so this doesn't corrupt game start.)
- **`readyPlayers` grows unbounded** (`LobbyService.kt:118`). `computeIfAbsent` creates an entry for any `gameId` (including non-existent / abandoned games; even `isReady=false` leaves an empty set), and no cleanup path reclaims bogus entries.

A guard at the top of `setReadyState` — game exists, caller is a member (`findByGameIdAndUserId`), and status is `WAITING` — fixes both majors plus the toggle-after-start and orphaned-entry minors.

### Minor / nits

- `toggleReady` has no `IN_PROGRESS` guard, so it can re-create the map entry `startGame` just dropped and emit a stale roster into a running game.
- Post-`startGame` `readyPlayers.remove(gameId)` runs outside the lock (`LobbyService.kt:349-350`), leaving a small orphaned-entry race window.
- The two new SonarCloud issues are most likely the unchecked `message.payload as? Map<*, *>` cast (now a third copy — consider a small `payloadMap()` helper) and the hard `as LobbyRosterDto` cast in the toggleReady test (use the safe `as? ... ?: throw` form used elsewhere in the file).
- Issue #343's "reset ready state when a new player joins" is not implemented (join path doesn't clear `readyPlayers`). Harmless because `startGame` re-validates the full roster, but it diverges from the AC — implement it or amend the issue.
- `LobbyWebSocketControllerTest` still asserts `PlayerNotFoundException` from `leaveLobby`, which the service can no longer throw (plus a now-dead import) — replace with a `LobbyDeleted` case.
- Join path issues two `getLobbyRoster` DB transactions; `/app/lobby.ready` is missing from `OpenApiConfig`/protocol docs; the new exception file has no trailing newline; several `feat:` commits are actually fixes/tests.

### What's good

Start-gate integrity against spoofed userIds, thread-safe ready-set reads/writes (no `ConcurrentModificationException`), correct success-path cleanup ordering, the positive `startGame` tests correctly re-guarded with `setReadyState`, precise broadcast assertions, robust `Number.toInt()` payload parsing, and an accurate wire-contract KDoc.

Bottom line: the two functional blockers (production griefing/false-state on leave, and the debug/host start-liveness break) are the gating items; the two majors share one small guard. None of the fixes are large.

Moritz2k1:
Requested changes have been implemented

PR #355, opened 2026-06-06
273 implement shopping mall landmark effect

### LINKED ISSUE #273: Implement Shopping Mall Landmark Effect
Shopping Mall landmark has no effect. 
Should add +1 income to "Bread" and "Cup" establishment payouts for the owning player.
                                                                                                                                                                                                                                                                                                                  What to implement:
- In EarningsServiceImpl.kt, check if active player has Shopping Mall built                                                                                                                                                                                                                                     - Apply +1 bonus per activation to qualifying green/blue food cards

SoupCube:
### Summary
Added changes to the earning logic. It now accounts for whether a player has build their Shopping mall. 
If they have, the player gets +1 coin for every Red CUP and Green BREAD card that gets triggerd.

### Changes
- Added a map containing the build status of the shopping mall per player
- Green and Red earning functions get this map when called
- Red logic check for every card if it is a Cup card and if the player has build the shopping mall
- Green logic check for every card if it is a Bread card and if the player has build the shopping mall
- If both criteria hold, the logic ads one coin for each of those cards to the total
- Added tests to confirm additional earnings when the shopping mall is built


Moritz2k1 (commented):
Overall correct:

Few things you can improve:

Missing: negative case for unbuilt landmark.
Both the unit test and integration test only verify the bonus IS applied when built. There is no test that isBuilt = false yields no bonus. If someone breaks the ?.isBuilt == true guard (e.g., removes the null-safe   check), no test catches it.

EarningsServiceImplTest.kt: Add test — Shopping Mall initialized but not built → no bonus. Mock findByPlayerIdAndType returning PlayerLandmarkModel(isBuilt = false), assert normal income.                 

Missing: quantity > 1.
The wiki description specifically calls out 2 Bakeries yielding 4 coins (2 × (1+1)). No test exercises playerCard.quantity > 1 with Shopping Mall. The formula playerCard.quantity * (card.income + extra) looks right but is untested for the multi-card case.                                           

EarningsServiceImplTest.kt: Add test — 2 Bakeries with Shopping Mall → 4 coins, not 3.

Missing: non-active player with Shopping Mall and GREEN card. Green cards only fire for the active player. If P2 owns Shopping Mall and Bakery but P1 is rolling, P2 should get no bonus. This case is not tested.  

EarningsIntegrationTest.kt: Add test — non-active player with Shopping Mall + Bakery → no green bonus.

SoupCube:
@Moritz2k1 
I added the additional tests

Moritz2k1 (approved):
Requested changes have been implemented. LGTM

PR #356, opened 2026-06-06
Fix: Ready Sync in Lobby

### LINKED ISSUE #343: Implement player ready state for lobby

Moritz2k1:
This PR fixes the ready synchronization in the lobby screen between multiple clients, by introducing an isReady property in the LobbyRosterDTO, a new concurrentHashmap (readyPlayers) and a new REST endpoint. 
The whole thing gets saved in memory, instead of using a flag in the database for the time being. This should be sufficient though.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyWebSocketController.kt
@@ -186,6 +186,60 @@ class LobbyWebSocketController(
     }
 
     /**
+     * Handles a player's ready-state toggle in the lobby.
+     *
+     * Client sends a message to `/app/lobby.ready` with `gameId` and `isReady`
+     * in the payload. Server updates the in-memory ready set and broadcasts
+     * the full updated roster to `/topic/game/{gameId}` so every client
+     * re-renders with accurate per-player ready state.
+     */
+    @MessageMapping("/lobby.ready")
+    fun toggleReady(
+        @Payload message: WebSocketMessage,
+        headerAccessor: SimpMessageHeaderAccessor,
+    ) {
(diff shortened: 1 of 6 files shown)


FabianSolutions (approved):
LGTM.

PR #359, opened 2026-06-06
275 implement radio tower landmark effect

### LINKED ISSUE #275: Implement Radio Tower Landmark Effect
Radio Tower landmark has no effect. 
Should allow player to reroll dice once per turn.

What to implement:
- DB migration: add rerolledThisTurn BOOLEAN DEFAULT FALSE to Games table
- Expose reroll action (endpoint or flag) for player to consume
- Reset flag on END_TURN in GamePhaseService

SoupCube:
### Summary 
Added the effect of the Radio Tower. When built the the player has the chance to reroll their dice once per turn.

### Changes
- Added a database migration, so that the game table holds a `rerolledThisTurn` boolean
- adjusted all function headers using the game table to account for the table changes
- Added functions to update the `rerroledThisTurn field` in the database in `GameDao.kt`
- Added a reroll method to `DiceService.kt`
- Added the reroll request to the `GameController.kt`
- Adjusted old tests to accept the new game table
- Added new tests to check the reroll behaviour in `DiceServiceTests.kt`


Moritz2k1 (commented):
Three changes needed:                                                                                                                                                                                                                         
1. GameController.kt:321 — dispatch to correct service method                                                                                                                val result = if (rollType == REROLL)                                                                                                                                          
    diceService.rerollDice(request, rollingPlayer.id)                                                                                                                         
else                                                                                                                                                                          
    diceService.rollDice(request, rollingPlayer.id)                                                                                                                           

2. DiceService.kt:65 — move tryMarkRerolledThisTurn after requireTrainStation                                                                                                 
val diceCount = resolveDiceCount(request)                                                                                                                                     
validateDiceCount(diceCount)                                                                                                                                                  
if (diceCount == TWO_DICE_COUNT) {                                                                                                                                            
    requireTrainStation(rollingPlayerId)                                                                                                                                      
}                                                                                                                                                                             
if (!gameDao.tryMarkRerolledThisTurn(request.gameId)) {                                                                                                                       
    throw CustomWebSocketException("REROLL_ALREADY_USED", ...)                                                                                                                }
                                                                                                                                                                     
3. GamePhaseService.kt:79 — bake reset into advanceTurn (atomic), remove separate call                                                                                                                                                                                                                            In GameDao.advanceTurn, add it[Games.rerolledThisTurn] = false inside the existing UPDATE block. Then remove gameDao.updateRerolledThisTurn(gameId, false) from GamePhaseService.                                                                                                                                                             

Also fix GameController.kt:316: Enum<DiceRoll> → DiceRoll.                                                                                                                    

And update GameControllerTest reroll tests to stub/verify diceService.rerollDice instead of diceService.rollDice

eles17 (commented):
One more thing to add to Moritz’s points: the reroll write should also be atomic with the current phase/roll state.

Right now `rerollDice` checks `turnPhase == RESOLVE_EFFECTS`, but `tryMarkRerolledThisTurn` only checks `rerolled_this_turn = false`, and `updateAfterRoll` unconditionally writes the new roll + phase. If `resolveEffects` runs concurrently after the state check, reroll can still mark the flag and set the game back to `RESOLVE_EFFECTS` after effects were already resolved.

Please make the reroll persistence a single conditional DAO operation, e.g. only update when:
- game id matches
- `turn_phase = RESOLVE_EFFECTS`
- `last_dice_roll IS NOT NULL`
- `rerolled_this_turn = false`

and in that same update set the new dice roll and `rerolled_this_turn = true`.

eles17:
@SoupCube please resolve conflicts

eles17 (dismissed):
Additional to resolving conflicts, I would still hold off on merging this PR for now.

There are a few issues left:

1. `rerollDice` still persists the reroll in two separate DAO writes:
   - `tryMarkRerolledThisTurn(...)`
   - `updateAfterRoll(...)`

   This does not fully address the atomicity concern from the review. The reroll should be persisted as one conditional DAO update that checks:
   - game id matches
   - `turn_phase = RESOLVE_EFFECTS`
   - `last_dice_roll IS NOT NULL`
   - `rerolled_this_turn = false`

   and in the same update writes the new dice roll and sets `rerolled_this_turn = true`.

2. `rerollDiceShouldThrowWhenTryingToRollTwoDiceWithoutTrainStation` looks stale. Since `requireTrainStation(...)` now runs before marking the reroll as used, this test should not expect `tryMarkRerolledThisTurn(1)` to be called.

3. The unauthenticated controller test named `rollDice throws UNAUTHENTICATED...` now calls `controller.rerollDice(...)`, so the original `rollDice` unauthenticated path is no longer covered by that test.

After those are fixed and tests/CI pass, this should be much closer to merge-ready.

Moritz2k1:
Fixed the failing tests

SoupCube:
@Moritz2k1 Thank you, I think this branch is done then

valentinaschiavon99 (dismissed):
Not merge-ready yet.

One requested-change item is still present: the controller test named `rollDice throws UNAUTHENTICATED when accessor has no principal` now calls `controller.rerollDice(...)` and verifies `diceService.rerollDice(...)`. That leaves the original `rollDice` unauthenticated path uncovered and makes the test name misleading.

Please update the unauthenticated coverage so both endpoints are explicit:
- keep `rollDice throws UNAUTHENTICATED when accessor has no principal` calling `controller.rollDice(...)` and verifying `diceService.rollDice(...)` is not called
- add a separate `rerollDice throws UNAUTHENTICATED when accessor has no principal` test that calls `controller.rerollDice(...)` and verifies `diceService.rerollDice(...)` is not called

After that is fixed and a reviewer re-approves, this should be merge-ready.

Moritz2k1:
Implemented. Re-review please

Moritz2k1:
Let's merge this first, then I will fix the other open PR

valentinaschiavon99 (approved):
Approved.

I checked the latest PR state against Server `main` and the current Client `main`.

Backend side:
- The reroll persistence is now handled atomically through `tryRerollThisTurn(...)`.
- The existing dice roll flow remains unchanged.
- The new reroll flow reuses the existing `ROLL_DICE` broadcast type, so current clients can still parse dice results.

Client compatibility:
- Current Client `main` remains backward-compatible because it parses `ROLL_DICE` payloads through `payload.result`.
- The new `DICE_REROLLED` event value is additive and should not break existing parsing.
- Client work is still needed to expose the new `/app/game.rerollDice` destination and add UI/ViewModel handling for Radio Tower reroll.

The missing client support should be tracked separately in the Client repository. This PR is fine to merge from the server side.

PR #362, opened 2026-06-08
274 implement amusement park landmark effect

### LINKED ISSUE #274: Implement Amusement Park Landmark Effect
Amusement Park landmark has no effect.
Should allow player to gain an extra turn when rolling doubles once per round.
 
- Detect doubles after 2-dice roll in DiceService.kt
- Add game state flag so player can take an extra turn
- Broadcast extra turn eligibility to clients via GameController
- give the player a second turn when their first turn ends (if eligible)

### LINKED ISSUE #277: Implement Landmark Effects
All 4 landmarks exist in the DB schema and are purchasable, but none have working game logic. 
This epic covers implementing each landmark's effect.

Child Issues:                                                                                                                                                                                                                                                                                                    
- Issue 2: Implement Shopping Mall Landmark Effect
- Issue 3: Implement Amusement Park Landmark Effect                                                                                                                                                                                                                                                              
- Issue 4: Implement Radio Tower Landmark Effect   
- Issue 5: Complete Train Station Landmark Effect                                                                                                                                                                                                                                                                
                                                   
Scope:
- DiceService.kt — doubles detection, 2-dice gating                                                                                                                                                                                                                                                              
- EarningsServiceImpl.kt — income modifiers                                                                                                                                                                                                                                                                      
- GamePhaseService.kt — reroll flag reset on end turn                                                                                                                                                                                                                                                            
- GameController.kt — broadcast reroll eligibility                                                                                                                                                                                                                                                               
- DB migration — rerolledThisTurn flag on Games table                                                                                                                                                                                                                                                            
                                                       
Done when: All 4 landmarks have working effects and are covered by tests.

SoupCube:
### Summary
Added the effect of the Amusement Park. When built the the player gets a second turn if the roll identical numbers on 2 dice.

### Changes
- Added a database migration, so that the game table holds the playerID and Round of an extra turn
- adjusted all function headers using the game table to account for the table changes
- Added functions to update the `extra_turn_player_id` and `extra_turn_round_number` field in the database in GameDao.kt
-  Added some adjustments to the rollDice method in `DiceService.kt` to check for same numbers on dice roll
- Added a check for if the player has the Amusement park built to `DiceService.kt`
- Added an `extraTurnGranted` boolean to the `rollDice` broadcast in `GameController.kt`
- Adjusted old tests to accept the new game table
- Added new tests to check the extra turn behaviour, broadcast and Amusement Park condition


eles17 (dismissed):
Thanks for the implementation and tests. I think the current behavior doesn’t match issue #274 yet.

The issue asks for Amusement Park to let the player **roll again when rolling doubles** and to broadcast **reroll eligibility**. This PR currently grants an extra full turn only after the player finishes resolving effects, optionally buying/building, and ending their turn.

Concretely, after doubles are rolled, `DiceService` still records the roll and moves the game to `RESOLVE_EFFECTS`, so another immediate `rollDice` call is rejected because `turnPhase != ROLL_DICE` / `lastDiceRoll != null`. The same player is only kept as active later in `GamePhaseService.endTurn`.

Requested change: please model this as reroll eligibility / an extra roll, not as an extra post-buy turn, or clarify the intended game rule if the issue wording is wrong.

eles17:
Pleas also update description, because it says  “Added the effect of the **Radio Tower**”, instead of Amusement Park

SoupCube:
@eles17 In the game rules the amusement park grants a second turn, the raido tower allows rerolling once per turn 
"Amusement Park:
A player may roll 2 dice only if they have already built a [Train Station]. In case of a double the player receives a complete additional turn. They may choose again whether to roll 1 or 2 dice, receive income and may build an Establishment or Landmark. The Amusement Park never gives more than one extra turn."  - https://www.yucata.de/en/Rules/MachiKoro

Changed the issue description to match the game rule

eles17:
Thanks for clarifying, you’re right: Amusement Park grants a complete extra turn, not a reroll. I mixed that up with Radio Tower.

One thing still looks off though: the extra-turn marker is cleared when the extra turn starts. If the player rolls doubles again during that extra turn, they can receive another extra turn in the same round.

Since issue #274 and the linked rules say Amusement Park only grants one extra turn, could you please keep track that the player already received the Amusement Park extra turn for this round?

SoupCube:
This branch has the same `AdminSeederTest` error as branch https://github.com/SE2-Machi-Koro/Server/pull/359 had
Tests running fine when the file runs alone, but fail when all test files are executed

valentinaschiavon99:
@SoupCube im taking it over to fix it

kutschera13 (dismissed):
Good implementation overall — the double-roll detection, Amusement Park 
gate and broadcasts look correct. Tests cover the main paths well (96.2% coverage ✅).

**🔴 Critical — GameDao.removeExtraTurnMark**
The WHERE clause is too broad and can clear another player's mark unintentionally. 
I suggest tightening the condition:

Games.update({
  (Games.id eq gameId) and
  (Games.extraTurnPlayerId eq playerId) and
  (Games.extraTurnRoundNumber eq roundNumber)
}) {
  it[Games.extraTurnPlayerId] = null
  it[Games.extraTurnRoundNumber] = null
}
I would like this fixed before merging.

**🟡 Important — DiceService reroll diceCount integrity**
The reroll path reads diceCount again from the client request, but lastDiceCount 
is not persisted on the server. This lets a malicious client switch from 1→2 dice 
on a reroll and trigger the Amusement Park extra-turn incorrectly. 
I recommend persisting lastDiceCount in the Games table and using the server-side 
value on reroll.

**🟡 Important — GamePhaseService.advanceTurn**
I don't see a guaranteed removal of the extra-turn mark when it is consumed. 
This can leave stale marks in the DB. I suggest clearing the mark atomically 
when consumeExtra is true — either call removeExtraTurnMark there, or extend 
advanceTurn to clear it in the same DB update.

Moritz2k1 (dismissed):
Bug: rerollDice ignores Amusement Park:
                                                                                                                                    rerollDice (DiceService.kt:56-91) never calls markExtraTurnIfEligible or removeExtraTurnMark. The Amusement Park check only runs in rollDice (line 44). So:                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                   Scenario A — roll doubles → grant set → Radio Tower reroll → non-doubles result → grant NOT cleared → advanceTurn sees consumeExtra=true → player gets extra turn despite non-doubles final result.
                                                                                                                                                                                                                                                                                                                Scenario B — roll non-doubles → no grant → Radio Tower reroll → doubles result → markExtraTurnIfEligible never called → player misses deserved extra turn.
                                                                                                                                                                                                                                                                                                        Per the wiki rule, Amusement Park applies to the result the player ends up with — the final dice result after any reroll.                                                                                                                                                                                        

Two smaller issues worth noting:
                                                                                                                                                                                  
removeExtraTurnMark condition too broad (GameDao.kt:187-195): condition A is extraTurnRoundNumber.isNotNull() alone — this would clear ANY player's extra turn mark, not just the caller's. No real impact because active player checks constrain callers, but logically wrong SQL.

Moritz2k1:
I will briefly fix this myself

valentinaschiavon99:
> I will briefly fix this myself

fine by me,thank u

Moritz2k1:
Re-review please and then we merge it

valentinaschiavon99 (dismissed):
Findings
Blocking: extra-turn marker can keep the same player active indefinitely.
[GamePhaseService.kt L75-L82](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/service/GamePhaseService.kt#L75-L82) treats extraTurnPlayerId + extraTurnRoundNumber as a pending extra turn every time endTurn runs. Since the marker is not cleared or converted to a “consumed” state, the first extra turn can end and still satisfy consumeExtra, keeping the same currentTurnIndex again. If the player rolls doubles during that extra turn, [DiceService.kt L100-L102](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/service/DiceService.kt#L100-L102) also ignores the markExtraTurnIfEligible return value, so the response can still report extraTurnGranted=true even when no new grant was persisted.
Suggested fix: model pending vs consumed explicitly, for example extra_turn_pending_player_id/round plus a consumed record, or clear the pending grant atomically when consumed while retaining a separate “already used this round” guard.

Blocking: reroll dice count is still client-controlled.
[DiceService.kt L84-L90](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/service/DiceService.kt#L84-L90) resolves reroll dice count from the request again. A client can roll one die initially, then submit a reroll request with two dice and potentially trigger Amusement Park logic at [L100-L101](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/service/DiceService.kt#L100-L101).
Suggested fix: persist lastDiceCount with the initial roll and force rerolls to use that server-side value.

Moritz2k1:
Implemented requested changes

valentinaschiavon99 (dismissed):
Findings
Blocking: Amusement Park can still grant chained extra turns in the same round.
[GamePhaseService.kt](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/service/GamePhaseService.kt#L75-L87) now clears extraTurnPlayerId/extraTurnRoundNumber when the extra turn is consumed. That prevents the marker from being consumed repeatedly, but it also removes the only record that the player already received their Amusement Park extra turn this round. During the extra turn, if the same player rolls doubles again, [GameDao.markExtraTurnIfEligible](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/dao/GameDao.kt#L169-L184) sees extraTurnRoundNumber.isNull() and grants another extra turn. This violates the rule already discussed in the PR: Amusement Park must not grant more than one extra turn.
Suggested fix: separate “pending extra turn” from “already granted this round”, or add a consumed/used marker that is retained until the relevant round advances

valentinaschiavon99:
Findings
Blocking: reroll can broadcast extraTurnGranted=false while the player still has a pending extra turn.
In [DiceService.kt](https://github.com/SE2-Machi-Koro/Server/blob/274-implement-amusement-park-landmark-effect/src/main/kotlin/org/machikoro/server/service/DiceService.kt#L91), rerollDice uses the return value of markExtraTurnIfEligible directly as extraTurnGranted. If the initial roll was doubles and already created a pending Amusement Park extra turn, then a Radio Tower reroll that also results in doubles will call markExtraTurnIfEligible, which correctly returns false because the same player/round already has a pending grant. The marker stays pending, so the player will still receive the extra turn, but the reroll broadcast says extraTurnGranted=false.
Suggested fix: for reroll doubles, return true when an existing non-consumed marker already belongs to the same player and round, or change the DAO method to return the effective pending-extra-turn state rather than only “new row update happened.” Add a test for: initial doubles grant pending -> reroll doubles -> broadcast extraTurnGranted=true and keep pending grant. @Moritz2k1

valentinaschiavon99 (approved):
No remaining blocking findings from my side. The Amusement Park extra-turn lifecycle and reroll broadcast behavior are now addressed, including the pending-grant reroll-doubles case.

Good job, Moritz.

PR #379, opened 2026-06-14
Fix: User ID implementation with Bearer Token

### LINKED ISSUE #372: `POST /lobby/{gameId}/start` trusts client-supplied `requestingUserId` instead of the authenticated principal
## Summary

The REST endpoint [`LobbyController.startGame`](../src/main/kotlin/org/machikoro/server/controller/LobbyController.kt) takes the acting user's identity from a **client-controlled query parameter** (`requestingUserId`) rather than from the authenticated session token. This lets any authenticated user start a game "on behalf of" the lobby host, violating the API-security requirement that a player must not be able to make calls on behalf of another player.

## Location

`src/main/kotlin/org/machikoro/server/controller/LobbyController.kt:19-26`

@PostMapping("/{gameId}/start")
fun startGame(
    @PathVariable gameId: Int,
    @RequestParam requestingUserId: Int,   // <-- supplied by the caller, not derived from the token
): ResponseEntity<Any> {
    val state = lobbyService.startGame(gameId, requestingUserId)
    return ResponseEntity.ok(state)
}

## Impact

- The endpoint requires a valid Bearer token to reach (`anyRequest().authenticated()` in `SecurityConfig`), but the host check inside `lobbyService.startGame(gameId, requestingUserId)` compares against the **forgeable** `requestingUserId` value.

- Any authenticated user who knows (or guesses) the host's user ID can force-start that lobby's game by calling `POST /lobby/{gameId}/start?requestingUserId=<hostUserId>`.

- This is a broken-object-level-authorization / IDOR issue. Severity is limited (the only action is starting a game), but it directly contradicts the stated requirement: *"Endpoints must ensure that the player is authorised to call them (e.g. Player 1 must not make calls on behalf of Player 2)."*

## Why the WebSocket path is not affected

The primary start path, `GameController.startGame` (`@MessageMapping("/game.start")`), derives the requesting user ID from the authenticated STOMP session via `WebSocketConnectionTracker` and never reads it from the payload. Only this REST convenience endpoint is vulnerable.

## Proposed fix

Derive the user ID from the security context and drop the `requestingUserId` request parameter:

import org.machikoro.server.auth.UserPrincipal
import org.springframework.security.core.annotation.AuthenticationPrincipal

@PostMapping("/{gameId}/start")
fun startGame(
    @PathVariable gameId: Int,
    @AuthenticationPrincipal principal: UserPrincipal,
): ResponseEntity<Any> =
    ResponseEntity.ok(lobbyService.startGame(gameId, principal.userId))

(`TokenAuthFilter` already places a `UserPrincipal` in the `SecurityContext` for authenticated REST requests, so `@AuthenticationPrincipal` resolves it directly.)

## Acceptance criteria

- [ ] `POST /lobby/{gameId}/start` no longer accepts or reads a `requestingUserId` parameter.

- [ ] The acting user is resolved from the authenticated token only.

- [ ] A non-host authenticated user calling the endpoint receives the existing not-host rejection, even if they pass another user's ID.

- [ ] Test covering: authenticated non-host cannot start a game by passing the host's user ID.

## Notes / related

- While reviewing, also consider whether `/auth/logout` (invalidates whatever `sessionToken` is in the body) needs tightening — currently acceptable since possession of the token is the identity, but worth a second look.

- All turn-bound gameplay actions (`purchase`, `endTurn`, `rollDice`, `rerollDice`, `resolveEffects`, `advancePhase`) already enforce per-player authorization correctly via `GameStateGuard.ensureSenderIsActivePlayer`.

Moritz2k1:
Changes:

LobbyController.kt — replaced @RequestParam requestingUserId: Int with @AuthenticationPrincipal principal: UserPrincipal. User ID now comes from the Bearer token (placed by TokenAuthFilter), not from a caller-supplied param.

LobbyControllerTests.kt — updated existing tests to pass UserPrincipal directly, added non-host cannot start game even if they know the host user ID which verifies the IDOR is closed: the controller always passes principal.userId to the service, so a non-host gets the NotHostException regardless of what they send.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/LobbyController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/LobbyController.kt
@@ -1,7 +1,9 @@
 package org.machikoro.server.controller
 
+import org.machikoro.server.auth.UserPrincipal
 import org.machikoro.server.service.LobbyService
 import org.springframework.http.ResponseEntity
+import org.springframework.security.core.annotation.AuthenticationPrincipal
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestMapping
@@ -13,15 +15,15 @@ class LobbyController(private val lobbyService: LobbyService) {
 
     /**
      * HTTP convenience endpoint for starting a game.
-     * The [requestingUserId] must match the game's hostUserId.
+     * Caller must be the lobby host — identity is taken from the Bearer token, not a param.
      * The primary start path is the WebSocket handler [GameController.startGame].
      */
     @PostMapping("/{gameId}/start")
     fun startGame(
         @PathVariable gameId: Int,
-        @org.springframework.web.bind.annotation.RequestParam requestingUserId: Int,
+        @AuthenticationPrincipal principal: UserPrincipal,
     ): ResponseEntity<Any> {
-        val state = lobbyService.startGame(gameId, requestingUserId)
+        val state = lobbyService.startGame(gameId, principal.userId)
         return ResponseEntity.ok(state)
     }
 }
--- a/src/test/kotlin/org/machikoro/server/controller/LobbyControllerTests.kt
+++ b/src/test/kotlin/org/machikoro/server/controller/LobbyControllerTests.kt
@@ -3,10 +3,12 @@ package org.machikoro.server.controller
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
+import org.machikoro.server.auth.UserPrincipal
 import org.machikoro.server.domain.enums.GameStatus
 import org.machikoro.server.domain.enums.TurnPhase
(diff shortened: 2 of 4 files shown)


valentinaschiavon99 (approved):
Reviewed and approved. Closes the IDOR on `POST /lobby/{gameId}/start` by deriving the caller identity from the Bearer-token `UserPrincipal` instead of a forgeable `requestingUserId` param. Verified:

- `TokenAuthFilter` sets a `UserPrincipal` as the auth principal for REST, so `@AuthenticationPrincipal` resolves correctly; `anyRequest().authenticated()` rules out a null principal.
- Aligns the HTTP start path with the existing WebSocket path (`GameController` already used `principal.userId`); `LobbyService.startGame` signature unchanged, no other callers affected.
- New `WithUserPrincipal` test helper is the right pattern for `@WebMvcTest`. Compiled the branch and ran LobbyControllerTests + RestErrorHandlingRegressionTest — both green.

Non-blocking note: `/lobby/**` isn't CSRF-exempt, so a direct HTTP call would need a CSRF token — pre-existing, and WebSocket is the documented primary path.

Issue #175, opened 2026-05-01
Buying duplicate Cards Logic


SoupCube:
### Usere Story
As a player, I want to own multiple copies of the same green, red, and blue establishments so that I can stack their effects.

### Acceptance Criteria
- Player can acquire multiple copies of the same establishment (green, red, blue)
- Player cannot acquire multiple copies of the same purple establishment
- Duplicate cards are tracked correctly in the game state
- Effects from multiple copies stack correctly

### Notes
- Backend must support card quantities, not just unique instances
- Backend must enforce rule (not just UI)

#### Related Issues
- https://github.com/SE2-Machi-Koro/Client/issues/107
- https://github.com/SE2-Machi-Koro/Client/issues/101

Moritz2k1:
This has already been covered.

eles17:
Reopening thus is partly done

### Problem
The backend supports duplicate card quantities, but it currently allows players to buy multiple copies of purple establishments. Machi Koro rules allow duplicate green, red, and blue establishments, but purple establishments must be unique per player.

PR #380, opened 2026-06-14
378 align server state and realtime events with client reconnect and dice safeguards

### LINKED ISSUE #378: Align server state and realtime events with client reconnect and dice safeguards
## Context

Client #175 added defensive handling for stale navigation snapshots and endless dice animations. The client now prevents stale state from skipping screens, prevents dice animation from spinning forever, and ignores duplicate roll taps while rolling.

The server should still align its state and event behavior so the client does not need to recover from stale, ambiguous, or misleading messages.

## Server-side items to check/fix

- `/app/chat.addUser` should not trust stale client-provided `gameId` as a fallback for reconnect/lobby membership.
- Login/reconnect should not revive stale `WAITING` / `IN_PROGRESS` memberships from old abandoned lobbies/games.
- Logout/leave flows should clean or invalidate lobby/game membership where appropriate.
- `lobby.create` should establish real host membership immediately, instead of relying on later join/addUser behavior.
- Duplicate `lobby.create` / `lobby.join` should be idempotent or rejected with clear behavior.
- Dice roll completion should always emit a stable completed result/event shape, including later rounds such as round 4.
- Add regression coverage for round 4 dice completion and stale client `gameId` handling.
- Verify STOMP `CONNECTED` session id behavior matches the private queue subscription strategy.

## Expected outcome

The server should avoid producing stale or misleading navigation/gameplay state in the first place, while preserving the client safeguards added in client #175 as a defensive fallback.

## Related

- Client #175
- To Align with https://github.com/SE2-Machi-Koro/Client/issues/175
- Existing server alignment thread: #291

eles17:
## Summary

This PR aligns server-side lobby/reconnect and dice event behavior with the client safeguards from Client [#175](https://github.com/SE2-Machi-Koro/Client/pull/342).

### Reconnect and stale state

- `/app/chat.addUser` no longer treats the client-provided `gameId` as a fallback source of truth.
- Reconnect handling now only joins/syncs a game when the server can resolve an active in-progress membership for the authenticated user.
- This prevents stale navigation snapshots or old client payloads from recreating misleading lobby/game membership.

### Lobby membership

- `lobby.create` now creates real host membership immediately.
- The host is added to the lobby roster at creation time instead of relying on a later join/addUser flow.
- `lobby.create` is idempotent when the user already has a waiting lobby, returning the existing waiting lobby instead of creating duplicates.
- `lobby.join` continues to use the existing add-user path, preserving duplicate join behavior for existing members.

### WebSocket session tracking

- `lobby.create` and `lobby.join` now register the authenticated session with the connection tracker.
- This keeps later game actions, such as `game.start`, tied to server-established membership instead of stale client-supplied state.

### Logout cleanup

- Logout now clears the user’s waiting lobby memberships through the existing lobby leave cleanup path.
- In-progress game membership is preserved so legitimate reconnects still work.

### Dice completion events

- Dice roll completion broadcasts now include `roundNumber` alongside the completed result shape.
- This keeps later-round dice events, including round 4, explicit and stable for clients.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/controller/GameController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/GameController.kt
@@ -193,10 +193,10 @@ class GameController(
                     gameId = gameId,
                 )
             )
-        } catch (e: GameStartedException) {
+        } catch (_: GameStartedException) {
             // Already initialized — idempotent, only runs once per game
             logger.debug("enterGameScreen: game $gameId already in progress, skipping initialization")
-        } catch (e: NotHostException) {
+        } catch (_: NotHostException) {
             // Non-host entered screen — host will trigger initialization when they enter
             logger.debug("enterGameScreen: userId=${principal.userId} is not host of game $gameId, skipping")
         }
@@ -265,7 +265,7 @@ class GameController(
         when (val result = gamePhaseService.endTurn(request.gameId)) {
             is EndTurnOutcome.Continue -> {
                 logger.info("Ended turn for game ${request.gameId}, new phase ${result.nextPhase}")
-                broadcastPhase(request.gameId, "TURN_ENDED")
+                broadcastTurnEnded(request.gameId)
             }
             is EndTurnOutcome.Won -> {
                 logger.info("Game ${request.gameId} finished, winner=${result.winnerId}")
@@ -448,11 +448,10 @@ class GameController(
     }
 
     /**
-     * Broadcasts the new turn phase and the active player's user ID to all
-     * subscribers of the game topic. The [activePlayerId] identifies the user
-     * whose turn it is so clients can compare it against their own user ID.
+     * Broadcasts the new turn phase and active player's user ID to all
+     * subscribers of the game topic.
      */
-    private fun broadcastPhase(gameId: Int, event: String) {
+    private fun broadcastTurnEnded(gameId: Int) {
         val state = gameSyncService.buildSnapshot(gameId)
         messagingTemplate.convertAndSend(
             "/topic/game/$gameId",
@@ -461,7 +460,7 @@ class GameController(
                 sender = "server",
                 payload = buildGameActionPayload(
                     state = state,
-                    event = event,
+                    event = "TURN_ENDED",
                 ),
                 gameId = gameId,
             ),
--- a/src/main/kotlin/org/machikoro/server/controller/GameController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/GameController.kt
@@ -330,6 +330,7 @@ class GameController(
                         "total" to result.total,
                         "completed" to result.completed,
                         "extraTurnGranted" to result.extraTurnGranted,
+                        "roundNumber" to state.game.roundNumber,
                         "timestamp" to System.currentTimeMillis(),
                         "state" to state,
                     ),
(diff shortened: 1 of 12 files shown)


valentinaschiavon99 (dismissed):
## Summary

The reconnect, lobby-membership, logout, and dice changes are mostly sound, but there is one blocking regression: a lobby host who reconnects (network blip or backend restart) can no longer start the game. Details and a couple of minor items below.

## Blocking: lobby reconnect leaves the session unregistered, breaking `game.start`

`addUser` now trusts only the server-side in-progress mapping and drops the client-supplied `gameId` fallback:

```kotlin
val mappedInProgressGameId = gameSyncService.findActiveInProgressGameId(principal.userId)
var gameIdToJoin = mappedInProgressGameId   // previously: ?: message.gameId
```

`findActiveInProgressGameId` resolves to `PlayerDao.findActiveGameIdByUserId`, which filters on `Games.status eq GameStatus.IN_PROGRESS`. For a `WAITING` lobby it returns `null`, so `gameIdToJoin` is `null` and the entire block, including `connectionTracker.register(...)`, is skipped.

This breaks the client's reconnect recovery. The client re-sends `addUser` on every STOMP `CONNECTED` (including reconnects) and relies on it to re-register the session with the connection tracker. `game.start` resolves the requesting user exclusively through the tracker, not the authenticated principal:

```kotlin
val requestingUserId = sessionId?.let { connectionTracker.getUserId(it) }
if (requestingUserId == null) { /* rejected: "Unknown session - please reconnect" */ }
```

Failure path: host creates a lobby (session registered by the new `lobby.create` code), the STOMP session drops and reconnects under a new session id, the client re-sends `addUser`, the server skips registration because the lobby is `WAITING`, and the host's subsequent `game.start` is rejected with `UNKNOWN_SESSION`. Before this change the client `gameId` fallback re-registered the new session, so start worked. This is particularly relevant for the backend-restart-recovery scenario, which is exactly the case that triggers a reconnect.

Suggested fix: re-register the session against a server-resolved waiting membership (for example via `findWaitingMembershipByUserId` / `findCurrentMembershipByUserId`) rather than the client-supplied `gameId`, or have `game.start` fall back to the authenticated principal instead of relying solely on the connection tracker.

## Minor: top-level `roundNumber` is not consumed by the current client

The added top-level `roundNumber` on the `DICE_ROLLED` and `GAME_ACTION` payloads is harmless but currently unused. The client reads `roundNumber` from the nested `payload.state` snapshot, not from the top-level field, so the stated goal (clients not needing to inspect nested state) is already satisfied through the existing snapshot path. This is fine to keep if another client is expected to consume it, but it is dead weight for the current client.

## Minor: pre-existing test missing `@Test`

`rerollDice throws UNAUTHENTICATED when accessor has no principal` in `GameControllerTest.kt` has no `@Test` annotation and therefore never runs. This predates this PR (the new round-4 test was inserted directly above it), but it would be worth fixing while the file is being touched.

## Looks correct

- `lobby.create` idempotency plus the immediate host roster add is internally consistent: the host is added once at creation, and `addUser` no longer double-adds since it is skipped for `WAITING` lobbies.
- Logout cleanup via `leaveWaitingLobbiesForUser` correctly preserves in-progress memberships so legitimate reconnects still work.
- The dice payload shape (`turnPhase`, `completed`, `result`) matches what the client parses.

Requesting changes for the reconnect/start regression.

eles17 (commented):
Thanks for the fix. This looks good to merge now. @Moritz2k1 or @FabianSolutions could you pls review/ approve this. Thanks.

Moritz2k1 (approved):
LGTM

PR #416, opened 2026-06-22
Optimize Railway deployment: JVM heap, health check, and dynamic port

### LINKED ISSUE #415: Optimize Railway deployment: JVM heap, health check, and dynamic port
## Summary

While auditing `main` for Railway readiness, the app deploys and runs, but a few things are not optimized for the image Railway actually runs (the prebuilt GHCR image, `runtime-from-workspace` target — **not** `compose.yaml`). This issue tracks the gaps.

### 1. JVM is not memory-aware in the production image
`Dockerfile` entrypoint is bare:
```dockerfile
ENTRYPOINT ["java", "-jar", "app.jar"]
```
`compose.yaml` sets `JAVA_TOOL_OPTIONS=-XX:MaxRAMPercentage=75.0`, but Railway pulls the GHCR image and does **not** use `compose.yaml`, so that flag never applies in production. The JVM then defaults to ~25% of container RAM for heap — wasting most of the Railway plan's memory and risking premature OOM/restarts under load.

**Fix:** bake it into the runtime stage — `ENV JAVA_TOOL_OPTIONS="-XX:MaxRAMPercentage=75.0"` — or set it as a Railway service env var.

### 2. No `railway.json` / `railway.toml`
No Railway config file exists, so deploys rely on a plain TCP check and the default restart policy. `/actuator/health` is exposed and `curl` is already in the image, so a health check would catch failed-but-listening deploys:
```json
{ "deploy": { "healthcheckPath": "/actuator/health", "restartPolicyType": "ON_FAILURE" } }
```

### 3. Fixed port instead of Railway's `$PORT`
`application.properties` binds `server.port=${SERVER_PORT:8080}`. This only works because Railway's target port is pinned to 8080. If Railway ever assigns a dynamic `$PORT`, the app won't bind.

**Fix:** `server.port=${PORT:${SERVER_PORT:8080}}`

## Minor / cleanup

- **`spring.docker.compose.enabled=true`** is the default in `application.properties`. Harmless in the prod jar (the `spring-boot-docker-compose` dependency is `developmentOnly`, so the autoconfig isn't on the runtime classpath), but the README still instructs setting `SPRING_DOCKER_COMPOSE_ENABLED=false` — confusing. Consider flipping the default to `false`.
- **`logging.level.org.flywaydb=DEBUG`** is permanently on — noisy production logs. Move to INFO.
- **Layered jar enabled but unused**: `bootJar { layered }` is on in `build.gradle.kts`, but the Dockerfile copies the whole jar instead of extracting layers — paying the build cost without the layer-caching benefit on rebuilds.

## Already good (no action)

- Multi-stage build, JRE-only runtime, non-root `app` user, `EXPOSE 8080`.
- Cache mount has the explicit `id` Railway's BuildKit requires.
- CI builds the jar once and packages multi-arch (amd64/arm64).
- DB config env-driven; `/actuator/health` permitted unauthenticated for probes; debug endpoints gated behind `DEBUG_ENABLED`.

valentinaschiavon99:
Closes #415.

This PR addresses **all** items from #415 — the three Railway-readiness fixes and the three cleanups.

## Railway readiness

- **JVM heap (Dockerfile)** — `ENV JAVA_TOOL_OPTIONS="-XX:MaxRAMPercentage=75.0"` on both runtime stages. The default container heap is only ~25% of the limit, leaving most of the platform's RAM unused.
- **Dynamic port (`application.properties`)** — `server.port` now resolves `${PORT:${SERVER_PORT:8080}}`, honoring Railway's injected `PORT` with `SERVER_PORT` then `8080` as fallbacks.
- **Deploy config (`railway.toml`)** — codified in-repo: `DOCKERFILE` builder, `healthcheckPath=/actuator/health`, `ON_FAILURE` restart policy (10 retries). Previously lived only in the Railway dashboard.

## Cleanups

- **Layered Docker image (Dockerfile)** — explode the layered bootJar with the Spring Boot `tools` jarmode and copy layers in least-to-most-likely-to-change order (`dependencies` → `spring-boot-loader` → `snapshot-dependencies` → `application`), launching via `JarLauncher`. Docker now reuses the heavy dependency layer on code-only rebuilds, speeding up Railway builds.
- **Docker Compose flag (`application.properties`)** — `spring.docker.compose.enabled` is now env-driven (`${SPRING_DOCKER_COMPOSE_ENABLED:true}`): defaults to `true` for local dev convenience, production sets it `false`. Makes the override point explicit.
- **Flyway logging (`application.properties`)** — lowered from `DEBUG` to `INFO` to quiet production logs.

## README

Corrected the Deployment section: Railway **builds from the `Dockerfile`** (per `railway.toml`); it does **not** pull the GHCR image as previously documented. The GHCR publish workflow still runs in parallel for versioned/rollback images.

## Verification

- Built the `runtime-from-workspace` image locally and ran it: image builds clean, app boots via `JarLauncher` (`LaunchedClassLoader`), Tomcat starts on 8080, `JAVA_TOOL_OPTIONS` is picked up, and it runs as the non-root `app` user. (Boot proceeds to the DB connection, as expected without a database.)
- Other changes are config/docs only — no Kotlin touched.

## Reviewer notes

- `railway.toml` mirrors exactly what the Railway dashboard generated.
- Spring Boot 4.0 uses the `tools` jarmode (`extract --layers --launcher`) and the `org.springframework.boot.loader.launch.JarLauncher` entry point — both verified against the built jar.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,11 +27,17 @@ EXPOSE 8080
 FROM runtime-base AS runtime-from-workspace
 COPY --chown=app:app --from=packaged-jar /app.jar /app/app.jar
 USER app
+# Make the JVM container-memory-aware: default heap is only ~25% of the
+# container limit, which wastes most of the platform's RAM (e.g. on Railway).
+ENV JAVA_TOOL_OPTIONS="-XX:MaxRAMPercentage=75.0"
 ENTRYPOINT ["java", "-jar", "app.jar"]
 
 FROM runtime-base AS runtime-from-builder
 COPY --chown=app:app --from=builder /app/build/libs/*.jar /app/app.jar
 USER app
+# Make the JVM container-memory-aware: default heap is only ~25% of the
+# container limit, which wastes most of the platform's RAM (e.g. on Railway).
+ENV JAVA_TOOL_OPTIONS="-XX:MaxRAMPercentage=75.0"
 ENTRYPOINT ["java", "-jar", "app.jar"]
 
 FROM runtime-from-builder AS final
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,20 +24,38 @@ RUN apt-get update \
 RUN groupadd --system app && useradd --no-log-init --system --create-home --home-dir /app --gid app app
 EXPOSE 8080
 
+# Explode the layered bootJar so Docker can cache slow-changing layers
+# (dependencies) separately from fast-changing ones (application code).
+FROM runtime-base AS extract-from-workspace
+COPY --from=packaged-jar /app.jar /app/app.jar
+RUN java -Djarmode=tools -jar /app/app.jar extract --layers --launcher --destination /app/extracted
+
+FROM runtime-base AS extract-from-builder
+COPY --from=builder /app/build/libs/*.jar /app/app.jar
+RUN java -Djarmode=tools -jar /app/app.jar extract --layers --launcher --destination /app/extracted
+
 FROM runtime-base AS runtime-from-workspace
-COPY --chown=app:app --from=packaged-jar /app.jar /app/app.jar
-USER app
 # Make the JVM container-memory-aware: default heap is only ~25% of the
 # container limit, which wastes most of the platform's RAM (e.g. on Railway).
 ENV JAVA_TOOL_OPTIONS="-XX:MaxRAMPercentage=75.0"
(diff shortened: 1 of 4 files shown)


eles17 (approved):
LGTM. Thanks for updating.

PR #426, opened 2026-06-28
fix(#425): enable STOMP broker heartbeats to keep idle WebSockets alive

### LINKED ISSUE #425: WebSocket idle connections dropped (~60s) causing reconnect churn / in-game lag — missing STOMP heartbeat
## Summary

WebSocket connections are silently dropped after ~60 seconds of inactivity, forcing clients to reconnect mid-game. The reconnect churn is experienced by players as lag, freezes, and "connection timeout". The server itself is healthy — it is not crashing, restarting, or running out of memory.

## Root cause

The STOMP message broker is configured **without heartbeats**, and the **native `/ws` endpoint** (used by the Android client) has **no heartbeat configured** — only the SockJS endpoint sets one.

`src/main/kotlin/org/machikoro/server/config/WebSocketConfig.kt`:
```kotlin
override fun configureMessageBroker(config: MessageBrokerRegistry) {
    config.enableSimpleBroker("/topic", "/queue")   // no heartbeat, no TaskScheduler
    config.setApplicationDestinationPrefixes("/app")
}
```

With no application-level heartbeat, an idle WebSocket carries no traffic during a player's think-time. Railway's edge proxy reaps the idle connection (~60s), and the client reconnects. During the reconnect, in-flight actions hang or are lost → perceived lag/timeout.

## Evidence (production logs, 2026-06-28)

The server ran continuously the whole window (no `Stopping Container`, no restart, no OOM), yet users `g` and `h` reconnect repeatedly, each reconnect preceded by a quiet/idle gap:

- `14:06:05` Resolved effects → quiet → `14:06:47` reconnect (~42s idle)
- `14:08:54` Resolved effects → quiet → `14:10:41` reconnect (~1m47s idle)

The idle gaps line up with the proxy's idle timeout, confirming idle-reaping rather than a crash or a client bug.

## Proposed fix

Enable application-level STOMP heartbeats so the socket stays alive even when no game messages are flowing:

```kotlin
override fun configureMessageBroker(config: MessageBrokerRegistry) {
    config.enableSimpleBroker("/topic", "/queue")
        .setHeartbeatValue(longArrayOf(10000, 10000))  // 10s each way
        .setTaskScheduler(brokerHeartbeatScheduler())   // REQUIRED — simple broker ignores heartbeats without a scheduler
    config.setApplicationDestinationPrefixes("/app")
}
```

- Add a `ThreadPoolTaskScheduler` bean for the broker heartbeats.
- Ensure the native `/ws` endpoint benefits (heartbeats are negotiated on the broker/inbound channel).
- **Client coordination required:** the Android client must advertise heartbeats in its CONNECT frame — both ends must agree, otherwise heartbeats stay disabled.

## Secondary observations (split out if preferred)

1. **Duplicate sessions per user.** Each disconnect logs `User Disconnected: <user>` three times at the same millisecond, and `joined the chat` fires multiple times per connect — each user appears to hold ~3 simultaneous WebSocket sessions. The client may be opening new connections without closing old ones (or re-subscribing on reconnect), which multiplies broadcasts and worsens lag.
2. **Noisy error logging.** `CustomWebSocketException: It is not your turn` (`GameStateGuard.kt:82` → `GameController.endTurn` `GameController.kt:266`) is logged at **ERROR with a full stack trace**. This is a normal validation rejection (possibly a stale action replayed after a reconnect) and should be logged at WARN/DEBUG without a stack trace so genuine errors aren't buried.

## Acceptance criteria

- [ ] A player can stay idle (thinking) for several minutes without the WebSocket being dropped.
- [ ] No reconnect churn in logs during normal idle periods.
- [ ] Broker heartbeats configured with a `TaskScheduler`; native `/ws` endpoint covered.
- [ ] Client negotiates heartbeats in the CONNECT frame.

valentinaschiavon99:
## What & why

Players reported lag, freezes, and connection timeouts during games. Production logs (2026-06-28) show the server is **healthy** — no crashes, no restarts, no OOM — yet clients repeatedly disconnect and reconnect, and each reconnect is preceded by a quiet/idle gap:

- `14:06:05` Resolved effects → quiet → `14:06:47` reconnect (~42s idle)
- `14:08:54` Resolved effects → quiet → `14:10:41` reconnect (~1m47s idle)

**Root cause:** the simple STOMP broker was configured **without heartbeats**, so during a player's think-time no traffic flows over the WebSocket. The platform proxy (Railway's edge) reaps the idle connection at ~60s, and the client reconnects — experienced as lag/timeouts. The native `/ws` endpoint (used by the Android client) had no keep-alive at all; only the SockJS endpoint set a heartbeat.

Fixes #425.

## Change

`WebSocketConfig.configureMessageBroker` now enables **bidirectional STOMP heartbeats** on the simple broker:

```kotlin
config.enableSimpleBroker("/topic", "/queue")
    .setHeartbeatValue(longArrayOf(HEARTBEAT_INTERVAL_MS, HEARTBEAT_INTERVAL_MS)) // 10s each way
    .setTaskScheduler(brokerHeartbeatScheduler())
```

- Heartbeat interval is **10s** in both directions (well under the proxy's ~60s idle window).
- The simple broker only honours heartbeats when a `TaskScheduler` is supplied, so a dedicated single-thread `ThreadPoolTaskScheduler` is registered as a `@Bean` (`ws-heartbeat-*`), giving it proper lifecycle management with the application context.
- Heartbeats are negotiated on the broker, so the native `/ws` endpoint is now covered too.

## Client coordination required ⚠️

Heartbeats are negotiated — **both ends must agree** or they stay disabled. The Android/STOMP client must advertise heartbeats in its CONNECT frame (e.g. `heart-beat: 10000,10000`). Without that, the server still won't emit keep-alives to a client that requests `0,0`. This needs a matching change on the client side to fully resolve the symptom in production.

## Testing

- Updated `WebSocketConfigTests`:
  - `configureMessageBrokerShouldEnableExpectedDestinations` now asserts the broker is configured with a `TaskScheduler` and a `10s,10s` heartbeat value.
  - Added `brokerHeartbeatSchedulerShouldBeInitialized` to confirm the scheduler bean is initialized and accepts work.
- `./gradlew test --tests "org.machikoro.server.config.WebSocketConfigTests"` → **BUILD SUCCESSFUL**.

## Out of scope (tracked in #425)

Two secondary observations from the logs are intentionally **not** addressed here to keep the PR focused:

1. **Duplicate sessions per user** — each disconnect logs the user ~3× at the same millisecond, suggesting the client opens connections without closing old ones / re-subscribes on reconnect. Likely a client-side fix.
2. **Noisy error logging** — `CustomWebSocketException: It is not your turn` (`GameStateGuard.kt:82` → `GameController.kt:266`) is logged at ERROR with a full stack trace; it's a normal validation rejection and should be WARN/DEBUG without a stack trace.

## Note on the earlier red herring

The June 27 logs showed a container living only ~26s before `Stopping Container` — that was a **one-off redeploy**, not a restart loop, and not the cause of the ongoing lag. The June 28 logs (continuous uptime + reconnect churn) confirmed the heartbeat issue instead.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/config/WebSocketConfig.kt
+++ b/src/main/kotlin/org/machikoro/server/config/WebSocketConfig.kt
@@ -60,7 +60,6 @@ class WebSocketConfig(
         ThreadPoolTaskScheduler().apply {
             poolSize = 1
             setThreadNamePrefix("ws-heartbeat-")
-            initialize()
         }
 
     /**
--- a/src/main/kotlin/org/machikoro/server/config/WebSocketConfig.kt
+++ b/src/main/kotlin/org/machikoro/server/config/WebSocketConfig.kt
@@ -2,9 +2,12 @@ package org.machikoro.server.config
 
 import org.machikoro.server.auth.StompAuthChannelInterceptor
 import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.messaging.simp.config.ChannelRegistration
 import org.springframework.messaging.simp.config.MessageBrokerRegistry
+import org.springframework.scheduling.TaskScheduler
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler
 import org.springframework.web.socket.config.annotation.EnableWebSocketMessageBroker
 import org.springframework.web.socket.config.annotation.StompEndpointRegistry
 import org.springframework.web.socket.config.annotation.WebSocketMessageBrokerConfigurer
@@ -18,6 +21,14 @@ class WebSocketConfig(
     companion object {
         const val NATIVE_WS_ENDPOINT = "/ws"
         const val SOCKJS_WS_ENDPOINT = "/ws-sockjs"
+
+        /**
+         * Interval (ms) the broker uses for STOMP heartbeats in both directions.
+         * Heartbeats keep otherwise-idle WebSocket connections alive so platform
+         * proxies (e.g. Railway's edge) do not reap them during a player's
+         * think-time, which previously forced reconnects mid-game.
+         */
+        const val HEARTBEAT_INTERVAL_MS = 10_000L
     }
 
     @Value("\${websocket.allowed-origins:http://localhost:8080,http://localhost:3000}")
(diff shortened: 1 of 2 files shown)


eles17 (dismissed):
Thanks for fix. Before merging, please remove the explicit `initialize()` call from `brokerHeartbeatScheduler()`. Since this scheduler is a Spring bean, Spring invokes `afterPropertiesSet()`, which initializes it again and replaces the first executor.

Please also update the scheduler test so it does not leave a running scheduler behind—either test it through the Spring context or explicitly call `destroy()` afterward.

eles17 (approved):
Looks ready to merge now. Approved

PR #428, opened 2026-06-28
fix(#427): register WebSocket exception handler as @ControllerAdvice

### LINKED ISSUE #427: Global WebSocket exception handler not applied — @Controller should be @ControllerAdvice (errors logged at ERROR, never sent to client)
## Summary

`GlobalWebSocketExceptionHandler` is annotated with `@Controller` instead of `@ControllerAdvice`, so its `@MessageExceptionHandler` methods are **never applied** to exceptions thrown by other controllers (e.g. `GameController`). As a result:

1. Domain `CustomWebSocketException`s (e.g. `NOT_YOUR_TURN`) are treated as **unhandled** by Spring's messaging layer and logged at **ERROR with a full stack trace**, instead of the intended WARN-without-trace.
2. The `@SendToUser("/queue/errors")` error reply is **never delivered to the client**, so the client receives no structured error for these conditions — a functional bug, not just noisy logs.

## Why

In Spring's messaging layer (`SimpAnnotationMethodMessageHandler`), `@MessageExceptionHandler` methods only handle exceptions thrown by `@MessageMapping` methods **within the same bean**, unless the bean is a `@ControllerAdvice` (which registers the handlers globally). `GlobalWebSocketExceptionHandler` is a `@Controller` with no `@MessageMapping` methods of its own, so its handlers currently catch nothing.

The REST sibling `RestExceptionHandler` is correctly `@RestControllerAdvice` (global); the WebSocket handler was meant to mirror it but used `@Controller`.

## Evidence (production logs, 2026-06-28)

A player ending their turn out of order throws `CustomWebSocketException("It is not your turn")` from `GameStateGuard.kt:82` (via `GameController.endTurn` → `requireActivePlayer`). The logs show Spring's fallback, **not** the handler's WARN:

```
ERROR ... .WebSocketAnnotationMethodMessageHandler : Unhandled exception from message handler method
org.machikoro.server.exception.CustomWebSocketException: It is not your turn
    at org.machikoro.server.service.GameStateGuard.ensureSenderIsActivePlayer(GameStateGuard.kt:82)
    at org.machikoro.server.controller.GameController.requireActivePlayer(GameController.kt:594)
    at org.machikoro.server.controller.GameController.endTurn(GameController.kt:266)
    ...
```

The expected `WARN "Handled WebSocket exception [NOT_YOUR_TURN]: It is not your turn"` line from `GlobalWebSocketExceptionHandler` never appears, confirming the handler is not invoked.

## Proposed fix

Annotate `GlobalWebSocketExceptionHandler` with `@ControllerAdvice` so its `@MessageExceptionHandler` methods apply globally to all `@MessageMapping` handlers. After the change:

- `CustomWebSocketException` → handled by `handleCustomException` → logged at WARN (no stack trace) and the error DTO is delivered to the client on `/queue/errors`.
- The generic `Exception` handler continues to log at ERROR (appropriate).

`src/main/kotlin/org/machikoro/server/handler/GlobalWebSocketExceptionHandler.kt`:
```kotlin
@ControllerAdvice   // was @Controller
class GlobalWebSocketExceptionHandler { ... }
```

## Acceptance criteria

- [ ] A `CustomWebSocketException` thrown by a `@MessageMapping` handler is caught by `GlobalWebSocketExceptionHandler`.
- [ ] It is logged at WARN without a stack trace (no more "Unhandled exception from message handler method" ERROR for these).
- [ ] The error DTO is delivered to the originating client on `/queue/errors`.
- [ ] Test coverage proves the advice is applied across controllers (e.g. an integration test exercising `NOT_YOUR_TURN`).

## Notes

This is independent of #425 (WebSocket heartbeats). Discovered while investigating the secondary log observations on PR #426.

valentinaschiavon99:
## What & why

WebSocket domain errors (e.g. `NOT_YOUR_TURN`) were both **mislogged** and **never delivered to the client**. Two distinct root causes, fixed here in sequence:

### 1. The handler wasn't wired in (`@Controller` → `@ControllerAdvice`)

`GlobalWebSocketExceptionHandler` was `@Controller`. In Spring's messaging layer, `@MessageExceptionHandler` methods only handle exceptions from `@MessageMapping` methods **in the same bean** unless the bean is `@ControllerAdvice`. This bean has no `@MessageMapping` methods, so its handlers caught nothing — exceptions thrown by `GameController` etc. were treated as unhandled and logged at **ERROR with a full stack trace** by Spring's fallback (`WebSocketAnnotationMethodMessageHandler`). Mirrors the REST side, which already uses `@RestControllerAdvice`.

### 2. The reply was silently dropped (`@SendToUser` → session-scoped send)

Making the handler run surfaced a second bug: its `@SendToUser("/queue/errors")` reply went nowhere. `@SendToUser` resolves the target through Spring's `SimpUserRegistry`, which **this app does not populate** — it authenticates with a token in the STOMP CONNECT body, not Spring Security — so the resolver finds no sessions and drops the message. This is the same limitation `WebSocketController.publishSync` already works around for game-sync.

Fix: send the error frame to the session-scoped destination `/queue/errors-user{sessionId}` directly (the destination Spring rewrites a `/user/queue/errors` subscription to), reading the originating session id from the message. `SimpMessagingTemplate` is injected for this.

Fixes #427.

## Evidence (production logs, 2026-06-28)

```
ERROR ... .WebSocketAnnotationMethodMessageHandler : Unhandled exception from message handler method
org.machikoro.server.exception.CustomWebSocketException: It is not your turn
    at org.machikoro.server.service.GameStateGuard.ensureSenderIsActivePlayer(GameStateGuard.kt:82)
    at org.machikoro.server.controller.GameController.endTurn(GameController.kt:266)
```
During development I confirmed the fix end-to-end: with `@ControllerAdvice` the handler logs `WARN "Handled WebSocket exception [NOT_YOUR_TURN]"`, and with the session-scoped send the integration test receives the error frame on `/user/queue/errors` (both previously failed/were silent).

## Testing

- **New end-to-end integration test** `WebSocketExceptionAdviceIntegrationTest`: drives a real `WebSocketStompClient`, triggers `NOT_YOUR_TURN` via `/app/game.endTurn`, and asserts the `WebSocketErrorDto` (`code=NOT_YOUR_TURN`, `message="It is not your turn"`) arrives on the caller's `/user/queue/errors`. This test fails on `main` (no frame delivered) and passes here.
- **Handler unit tests** updated for the send-based behavior: assert the DTO is sent to `/queue/errors-user{sessionId}`, that the internal-error message is sanitized (no DB URL / password leak), that a missing session id is a no-op, and a regression guard asserting the `@ControllerAdvice` annotation.
- Re-ran `GameWebSocketBroadcastIntegrationTest` to confirm the now-active global handler does not disturb the existing purchase-rejection-to-topic broadcast path. All green.

## Scope

- Independent of #425 / PR #426 (WebSocket heartbeats); branched off `main`.
- The other secondary observation from PR #426 — duplicate sessions per user — is primarily client-side and intentionally **not** addressed here.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/handler/GlobalWebSocketExceptionHandler.kt
+++ b/src/main/kotlin/org/machikoro/server/handler/GlobalWebSocketExceptionHandler.kt
@@ -4,7 +4,8 @@ import org.machikoro.server.dto.WebSocketErrorDto
 import org.machikoro.server.exception.CustomWebSocketException
 import org.slf4j.LoggerFactory
 import org.springframework.messaging.handler.annotation.MessageExceptionHandler
-import org.springframework.messaging.simp.annotation.SendToUser
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor
+import org.springframework.messaging.simp.SimpMessagingTemplate
 import org.springframework.web.bind.annotation.ControllerAdvice
 
 /**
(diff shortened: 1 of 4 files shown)


eles17 (approved):
LGTM. Approved

PR #431, opened 2026-06-29
fix: Jackson strips the prefix from isBuilt()

### LINKED ISSUE #430: bug: when radio tower is built the game is stuck
After building the radio tower the game gets stuck in the resolve effects phase

Moritz2k1:


valentinaschiavon99 (dismissed):
Good catch on the `isBuilt` fix — this resolves the radio-stuck bug. Requesting changes because the same latent bug exists elsewhere and the root cause isn't addressed:

**`LobbyRosterPlayerDto.isReady` (src/main/kotlin/org/machikoro/server/dto/LobbyRosterDto.kt:14) has the identical bug.** It's serialized to lobby clients and hits the same Jackson `is`-prefix stripping, so it goes out on the wire as `ready`, not `isReady`. If the Android client reads `isReady`, the lobby ready-state is silently always-false/absent — exactly the failure this PR fixes for landmarks. (`isAdmin` in UserModel.kt:15 is safe — it's `@JsonIgnore`.)

**This is a per-field bandaid for a global serialization-config problem.** There's no central ObjectMapper/module configuration ensuring Kotlin `isX` property names are preserved, so each boolean field re-discovers this bug in production. Please either:
- fix it centrally — ensure the Kotlin module's property-name handling is actually applied (or configure the mapper's accessor naming) so every `isX` boolean serializes correctly without per-field annotations, **or**
- at minimum, also annotate `isReady` here and add a note/test so the next `isX` field doesn't silently regress.

Nit: the file is also missing a trailing newline at EOF (pre-existing, but the diff preserves it).

Moritz2k1:
Implemented

valentinaschiavon99 (dismissed):
## Code Review: PR #431 - Jackson strips the prefix from isBuilt()

### Finding: Test's JsonMapper doesn't include KotlinModule

**Severity**: Medium  
**File**: `src/test/kotlin/org/machikoro/server/config/WebSocketConfigTests.kt` · Line 20

#### Problem
The test creates a JsonMapper without KotlinModule:
```kotlin
private val config = WebSocketConfig(authInterceptor, JsonMapper.builder().build()).apply {
```

This mapper will **strip "is" prefixes** from boolean properties, which defeats the purpose of this PR. The test will pass, but it won't actually validate that boolean names are preserved during serialization. If production's JsonMapper configuration fails, the test won't catch it.

#### What will happen
- Test passes ✅
- But boolean properties serialize as `"built"` and `"ready"` instead of `"isBuilt"` and `"isReady"` ❌
- Production bug goes undetected

#### How to fix (pick one)

**Option A: Inject Spring Boot's auto-configured JsonMapper** (Recommended)
```kotlin
@Autowired
private lateinit var springJsonMapper: JsonMapper

private val config = WebSocketConfig(authInterceptor, springJsonMapper).apply {
    allowedOrigins = "http://localhost:8080,http://localhost:3000"
}
```
✅ Tests production's actual setup  
✅ Guaranteed to have KotlinModule  
✅ Catches real failures

**Option B: Explicitly add KotlinModule to the builder**
```kotlin
private val config = WebSocketConfig(
    authInterceptor,
    JsonMapper.builder()
        .addModule(com.fasterxml.jackson.module.kotlin.KotlinModule())
        .build()
).apply {
    allowedOrigins = "http://localhost:8080,http://localhost:3000"
}
```
✅ Clear what's needed  
⚠️ Duplication (also configured globally)

---

### Review Summary

The PR itself is well-designed:
- ✅ Global fix (configureMessageConverters) correctly implemented
- ✅ Field-level annotations (@get:JsonProperty) provide good defense-in-depth
- ✅ Spring Boot 4.0.3 properly auto-configures JsonMapper with KotlinModule
- ✅ No breaking changes

**Ready to merge once the test is updated to validate the fix.**
@Moritz2k1

valentinaschiavon99 (dismissed):
## Code Review: PR #431 - Jackson strips the prefix from isBuilt()

### Finding: Test's JsonMapper doesn't include KotlinModule

**Severity**: Medium  
**File**: `src/test/kotlin/org/machikoro/server/config/WebSocketConfigTests.kt` · Line 20

#### Problem
The test creates a JsonMapper without KotlinModule:
```kotlin
private val config = WebSocketConfig(authInterceptor, JsonMapper.builder().build()).apply {
```

This mapper will **strip "is" prefixes** from boolean properties, which defeats the purpose of this PR. The test will pass, but it won't actually validate that boolean names are preserved during serialization. If production's JsonMapper configuration fails, the test won't catch it.

#### What will happen
- Test passes ✅
- But boolean properties serialize as `"built"` and `"ready"` instead of `"isBuilt"` and `"isReady"` ❌
- Production bug goes undetected

#### How to fix (pick one)

**Option A: Inject Spring Boot's auto-configured JsonMapper** (Recommended)
```kotlin
@Autowired
private lateinit var springJsonMapper: JsonMapper

private val config = WebSocketConfig(authInterceptor, springJsonMapper).apply {
    allowedOrigins = "http://localhost:8080,http://localhost:3000"
}
```
✅ Tests production's actual setup  
✅ Guaranteed to have KotlinModule  
✅ Catches real failures

**Option B: Explicitly add KotlinModule to the builder**
```kotlin
private val config = WebSocketConfig(
    authInterceptor,
    JsonMapper.builder()
        .addModule(com.fasterxml.jackson.module.kotlin.KotlinModule())
        .build()
).apply {
    allowedOrigins = "http://localhost:8080,http://localhost:3000"
}
```
✅ Clear what's needed  
⚠️ Duplication (also configured globally)

---

### Review Summary

The PR itself is well-designed:
- ✅ Global fix (configureMessageConverters) correctly implemented
- ✅ Field-level annotations (@get:JsonProperty) provide good defense-in-depth
- ✅ Spring Boot 4.0.3 properly auto-configures JsonMapper with KotlinModule
- ✅ No breaking changes

**Ready to merge once the test is updated to validate the fix.**
@Moritz2k1

Moritz2k1:
Implemented option B. Can we merge this now?

valentinaschiavon99 (dismissed):
Thanks for tracking this down — the root-cause analysis is right: the STOMP broker built its own default JSON converter with a plain mapper lacking `KotlinModule`, unlike the HTTP path. The core fix (injecting Boot's `JsonMapper` and wiring it via `configureMessageConverters`) is correct. A few things to address before merge:

**1. The actual fix has no regression test.**
`WebSocketConfigTests` was only edited so the constructor call compiles — nothing asserts that `configureMessageConverters` registers a `KotlinModule`-enabled converter, or that `isBuilt`/`isReady` serialize *with* their prefix. A future refactor that drops the override or the `JsonMapper` injection would silently reintroduce this exact bug with the suite still green. Please add a test that serializes `PlayerLandmarkModel` / `LobbyRosterPlayerDto` through the configured converter and asserts the JSON contains `"isBuilt"` / `"isReady"`.

**2. The per-DTO `@get:JsonProperty` annotations are redundant and inconsistently applied.**
Once the converter fix is in place, the prefix is preserved globally for every WebSocket message, so the annotations on `PlayerLandmarkModel` and `LobbyRosterPlayerDto` are belt-and-suspenders — and they're applied to only 2 of the codebase's `is`-prefixed types. The inconsistency actually masks regressions: if the global fix is ever removed, these two keep working while every other `is*` DTO breaks. Prefer relying on the converter fix alone (drop the annotations), or add a comment explaining why both layers are intentional.

**3. Missing newline at end of file** in `WebSocketConfig.kt`, `PlayerLandmarkModel.kt`, and `LobbyRosterDto.kt` (`\ No newline at end of file` in all three hunks). Please restore the trailing newline.

Moritz2k1:
Implemented

eles17 (approved):
LGTM. Approved.

eles17 (approved):
aaprove

PR #438, opened 2026-06-30
fix: make WinConditionService.detectWinner phase-agnostic (#437)

### LINKED ISSUE #437: WinConditionService.detectWinner unnecessarily enforces END_TURN phase
## Summary

`WinConditionService.detectWinner` unnecessarily enforces the `END_TURN` phase, throwing `NOT_END_TURN_PHASE` if called outside it. The guard is redundant and will break any future non-`END_TURN` call path.

## Location

`WinConditionService.kt:28–32`

## Root cause

Throws `NOT_END_TURN_PHASE` if called outside `END_TURN`. Since it is always called after transitioning to `END_TURN` anyway, the guard is redundant and will break any future non-`END_TURN` call path (e.g. a debug or mid-turn check).

## Fix

Remove the phase guard, or relax it to an assertion rather than a thrown exception.

---

valentinaschiavon99:
## Summary

`WinConditionService.detectWinner` unnecessarily enforced the `END_TURN` turn phase, throwing `CustomWebSocketException("NOT_END_TURN_PHASE")` whenever it was called in any other phase. That guard was redundant — a winner is simply whoever has built all landmarks, which is independent of the current `TurnPhase` — and `detectWinner` is already invoked right after transitioning to `END_TURN`. Worse, the guard actively broke any future non-`END_TURN` call path (e.g. a debug or mid-turn win check).

This PR makes `detectWinner` **phase-agnostic** by removing the guard.

Closes #437

## Changes

- **`WinConditionService.kt`**
  - Removed the `turnPhase != END_TURN` check and the `NOT_END_TURN_PHASE` exception.
  - The game is still looked up so a missing game continues to throw `GameNotFoundException`.
  - Updated the KDoc to document that the check is phase-agnostic.
  - Dropped the now-unused `CustomWebSocketException` import.
- **`WinConditionServiceTest.kt`**
  - Replaced `detectWinner throws if not in END_TURN phase` with `detectWinner is phase-agnostic and detects a winner outside END_TURN`, which sets the game to `BUY_OR_BUILD` and asserts the winner is still returned.
  - Dropped the now-unused `CustomWebSocketException` import.

## Behaviour

| Scenario | Before | After |
| --- | --- | --- |
| All landmarks built, phase `END_TURN` | returns winner | returns winner |
| All landmarks built, phase `BUY_OR_BUILD` | throws `NOT_END_TURN_PHASE` | returns winner |
| No player has won | returns `null` | returns `null` |
| Game not found | throws `GameNotFoundException` | throws `GameNotFoundException` |

The existing production caller, `GamePhaseService` (`detectWinner(gameId)?.let { ... }`), is unaffected since it always calls after the `END_TURN` transition.

## Testing

- `./gradlew test --tests WinConditionServiceTest --tests GamePhaseServiceTest` — **BUILD SUCCESSFUL**

### DIFF
--- a/src/main/kotlin/org/machikoro/server/service/WinConditionService.kt
+++ b/src/main/kotlin/org/machikoro/server/service/WinConditionService.kt
@@ -5,7 +5,6 @@ import org.machikoro.server.dao.PlayerDao
 import org.machikoro.server.dao.PlayerLandmarkDao
 import org.machikoro.server.domain.enums.TurnPhase
 import org.machikoro.server.domain.models.PlayerModel
-import org.machikoro.server.exception.CustomWebSocketException
 import org.machikoro.server.exception.GameNotFoundException
 import org.springframework.stereotype.Service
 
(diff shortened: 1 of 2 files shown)


eles17 (approved):
LGTM. Approved

PR #439, opened 2026-06-30
refactor: remove dead atomicity-bypassing GameDao methods

### LINKED ISSUE #435: GameDao.updateAfterRoll and updateRerolledThisTurn are dead production code bypassing atomicity guards
## Summary

`GameDao.updateAfterRoll` and `GameDao.updateRerolledThisTurn` are dead production code — they are only called from tests, and they bypass the atomicity guards.

## Location

`GameDao.kt:146, 215`

## Root cause

Both methods are only called from tests. They bypass the atomicity guards (`tryRecordDiceRoll` / `tryRerollThisTurn`) and risk accidental misuse if called from production code.

## Fix

Remove them from production code or restrict them to test scope only.

---

valentinaschiavon99:
## Summary

Closes #435.

`GameDao.updateAfterRoll` and `GameDao.updateRerolledThisTurn` were **dead production code** — neither was called anywhere outside of tests, and both wrote game state with an unconditional `UPDATE ... WHERE id = ?` that bypassed the atomicity guards used everywhere else in the DAO:

- `updateAfterRoll` set `lastDiceRoll` + `turnPhase` directly, sidestepping `tryRecordDiceRoll`, which only allows the `ROLL_DICE → RESOLVE_EFFECTS` transition once (`turnPhase = ROLL_DICE AND lastDiceRoll IS NULL`).
- `updateRerolledThisTurn` flipped `rerolledThisTurn` directly, sidestepping `tryRerollThisTurn`, which guards the one-reroll-per-turn rule.

Leaving them in the production DAO meant a future caller could accidentally bypass those concurrency guards and reintroduce the double-roll / double-reroll races the guarded methods exist to prevent.

## Changes

**Production (`GameDao.kt`)**
- Removed `updateAfterRoll(id, diceRoll, phase)`.
- Removed `updateRerolledThisTurn(id, rerolledThisTurn)`.

**Tests** — migrated every setup usage to the guarded production path so fixtures now exercise the real transition logic:
- `RESOLVE_EFFECTS` setup → `tryRecordDiceRoll(id, diceRoll, diceCount = 1)` (sets the dice roll and advances the phase exactly as production does).
- `BUY_OR_BUILD` setup → `tryRecordDiceRoll(...)` followed by `updateTurnPhase(id, BUY_OR_BUILD)`.
- Affected files: `GameDaoTest.kt`, `PurchaseServiceIntegrationTest.kt`, `GameSyncServiceIntegrationTest.kt`, `SpringContextRestartRecoveryIntegrationTest.kt`.
- Dropped the four tests in `GameDaoTest` that only existed to exercise the removed methods (`updateAfterRoll sets dice result and phase`, `updateAfterRoll throws when game does not exist`, `updateRerolledThisTurn changes rerolled flag when game exists`, `updateRerolledThisTurn throws when game does not exist`).

## Why this approach

Rather than restrict the methods to test scope, the test fixtures already had a guarded equivalent available (`tryRecordDiceRoll`), so the bypass could be removed entirely. Setups now go through the same code path production uses, which is both safer and more representative.

### DIFF
--- a/src/main/kotlin/org/machikoro/server/dao/GameDao.kt
+++ b/src/main/kotlin/org/machikoro/server/dao/GameDao.kt
@@ -141,17 +141,6 @@ class GameDao {
     }
 
     /**
-     * Updates game state after a dice roll
-     */
-    fun updateAfterRoll(id: Int, diceRoll: Int, phase: TurnPhase): Unit = transaction {
-        val updatedRows = Games.update({ Games.id eq id }) {
-            it[Games.lastDiceRoll] = diceRoll
-            it[Games.turnPhase] = phase
-        }
-        if (updatedRows == 0) throw GameNotFoundException("Game $id not found")
-    }
-
-    /**
      * Persists the only legal transition out of ROLL_DICE. The conditional
      * update prevents two simultaneous requests from recording separate rolls.
      */
@@ -207,16 +196,6 @@ class GameDao {
         }) {
             it[Games.turnPhase] = next
         } > 0
-    }
-
-    /**
-     * Updates the rerolled_this_turn flag for a game.
-     */
-    fun updateRerolledThisTurn(id: Int, rerolledThisTurn: Boolean): Unit = transaction {
-        val updatedRows = Games.update({ Games.id eq id }) {
-            it[Games.rerolledThisTurn] = rerolledThisTurn
-        }
-        if (updatedRows == 0) throw GameNotFoundException("Game $id not found")
     }
 
     /**
--- a/src/main/kotlin/org/machikoro/server/dao/GameDao.kt
+++ b/src/main/kotlin/org/machikoro/server/dao/GameDao.kt
@@ -206,6 +206,10 @@ class GameDao {
      *
      * On success, sets the new dice roll and marks rerolledThisTurn = true in a single transaction.
      * Returns true if the update succeeded, false if any condition failed.
+     *
+     * lastDiceCount is intentionally left untouched: a reroll replays the same number of dice as the
+     * initial roll, so the count recorded by [tryRecordDiceRoll] must be preserved (DiceService reads
+     * lastDiceCount at reroll time to know how many dice to roll).
      */
     fun tryRerollThisTurn(id: Int, newDiceRoll: Int): Boolean = transaction {
         Games.update({
@@ -216,6 +220,7 @@ class GameDao {
         }) {
             it[Games.lastDiceRoll] = newDiceRoll
             it[Games.rerolledThisTurn] = true
+            // lastDiceCount deliberately not updated — the reroll keeps the initial roll's dice count.
         } > 0
     }
 
--- a/src/test/kotlin/org/machikoro/server/dao/GameDaoTest.kt
+++ b/src/test/kotlin/org/machikoro/server/dao/GameDaoTest.kt
@@ -116,23 +116,6 @@ class GameDaoTest : AbstractDBSetup() {
     }
 
     @Test
-    fun `updateAfterRoll sets dice result and phase`() {
-        val id = gameDao.create(hostId)
-        gameDao.updateAfterRoll(id, diceRoll = 6, phase = TurnPhase.RESOLVE_EFFECTS)
-        val game = gameDao.findById(id)!!
-        assertEquals(6, game.lastDiceRoll)
-        assertEquals(TurnPhase.RESOLVE_EFFECTS, game.turnPhase)
-        assertFalse(game.hasPurchasedThisTurn)
-    }
-
-    @Test
-    fun `updateAfterRoll throws when game does not exist`() {
-        assertThrows<GameNotFoundException> {
-            gameDao.updateAfterRoll(999999, diceRoll = 6, phase = TurnPhase.RESOLVE_EFFECTS)
-        }
-    }
-
-    @Test
     fun `tryRecordDiceRoll records exactly one roll in a turn`() {
         val id = gameDao.create(hostId)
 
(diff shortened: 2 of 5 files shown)


valentinaschiavon99 (commented):
## ✅ Approve (posting as comment — GitHub blocks self-approval)

Clean, well-justified refactor. `updateAfterRoll` and `updateRerolledThisTurn` are genuinely dead production code, and both did unconditional `UPDATE ... WHERE id = ?` writes that bypassed the atomicity guards (`tryRecordDiceRoll` / `tryRerollThisTurn`). Removing them eliminates a real footgun.

**Correctness verified:**
- No production callers remain; only test setups referenced them.
- Every migrated `tryRecordDiceRoll` call is preceded by `create()` or `advanceTurn()`, both of which set `ROLL_DICE` + `lastDiceRoll = null` — so the guard's precondition holds at each call site.
- `RESOLVE_EFFECTS` setups: `tryRecordDiceRoll` lands on `RESOLVE_EFFECTS`, matching the old end state.
- `BUY_OR_BUILD` setups: `tryRecordDiceRoll` + `updateTurnPhase(BUY_OR_BUILD)` reproduces the old end state, and now also sets `lastDiceCount`.
- Dropping the four tests that only exercised the removed methods is correct.

**Minor (non-blocking):** the migrated setups ignore the `Boolean` return of `tryRecordDiceRoll`. The old `updateAfterRoll` threw on a missing game; the new path silently no-ops if its precondition ever fails, so a future fixture regression would surface as a confusing downstream assertion rather than at the setup line. Optional hardening: `check(gameDao.tryRecordDiceRoll(...)) { "fixture: expected ROLL_DICE" }`.

Safe to merge once CI is green.

valentinaschiavon99:
@Moritz2k1 check it and tell if we can merge it

Moritz2k1:
Review: refactor/435-remove-bypass-gamedao-methods
====================================================

The core premise is correct — updateAfterRoll and updateRerolledThisTurn were genuine
footguns and deleting them is the right call. The existing review already caught the most
important finding (unchecked returns), so this focuses on what it missed.


1. Unchecked return from tryRecordDiceRoll — confirmed, but the risk is higher than noted
------------------------------------------------------------------------------------------
Files: GameDaoTest.kt:174,338, GameSyncServiceIntegrationTest.kt:112,
       PurchaseServiceIntegrationTest.kt:204,219, SpringContextRestartRecoveryIntegrationTest.kt:139

All four call sites discard the Boolean. The existing review flagged this as "optional
hardening." It's a real fixture bug waiting to happen.

The precondition is: turnPhase = ROLL_DICE AND lastDiceRoll IS NULL.
SpringContextRestartRecoveryIntegrationTest uses a sharedGameId shared across @Test methods —
if test ordering or a prior test leaves the game in a non-ROLL_DICE state, tryRecordDiceRoll
silently no-ops and the test carries on with lastDiceRoll = null. The failure surfaces two
assertions later with a message that points nowhere near the setup. Minimum fix:

    check(gameDao.tryRecordDiceRoll(gameId, diceRoll = 8, diceCount = 2)) {
        "fixture: game not in ROLL_DICE"
    }


2. updateTurnPhase is still an unconditional bypass in the migrated BUY_OR_BUILD setups
-----------------------------------------------------------------------------------------
Files: GameSyncServiceIntegrationTest.kt:113, SpringContextRestartRecoveryIntegrationTest.kt:140

The migrated pattern is:
    gameDao.tryRecordDiceRoll(gameId, diceRoll = 8, diceCount = 2)  // guarded
    gameDao.updateTurnPhase(gameId, TurnPhase.BUY_OR_BUILD)          // unconditional

updateTurnPhase is the same shape as the removed methods:
    UPDATE games SET turn_phase = ? WHERE id = ?  — no state check.

The guarded equivalent already exists and is exactly what production uses:
    // EarningsServiceImpl.kt:279
    gameDao.tryTransitionPhase(gameId, TurnPhase.RESOLVE_EFFECTS, TurnPhase.BUY_OR_BUILD)

Using tryTransitionPhase here would complete the goal ("fixtures now exercise the real
transition logic") for the BUY_OR_BUILD path. As-is, the fixture bypasses the same kind of
guard the PR set out to close — just for a different transition.


3. tryRerollThisTurn doesn't update lastDiceCount — pre-existing, now more visible
------------------------------------------------------------------------------------
Files: GameDao.kt:210, DiceService.kt:75

DiceService reads lastDiceCount at reroll time to know how many dice to roll (:75).
tryRerollThisTurn updates lastDiceRoll but not lastDiceCount. That's intentional (same dice
count for the reroll), so it's correct — but now that fixtures set lastDiceCount via
tryRecordDiceRoll rather than bypassing it, this invariant is more load-bearing. Worth a
comment in tryRerollThisTurn that lastDiceCount is intentionally preserved from the initial
roll.


Summary
-------
Approve with fixes. Issue 1 is the most urgent (silent fixture failure in shared-state
tests), issue 2 is the one that actually contradicts the PR's stated goal, and issue 3 is
a note. None block correctness of the production code — all are in test setup.

valentinaschiavon99:
## Review

Tight, well-scoped refactor. Verified the removed-behavior, cross-file, and line-by-line angles:

- **No dangling callers** — grep across `src/` confirms nothing references the removed `updateAfterRoll` / `updateRerolledThisTurn`.
- **Fixture substitutions are equivalent** — `tryRecordDiceRoll` sets `lastDiceRoll`/`turnPhase = RESOLVE_EFFECTS` (the old target), and the BUY_OR_BUILD setups layer `tryTransitionPhase(RESOLVE_EFFECTS → BUY_OR_BUILD)` on top to reach the same end state. Each is wrapped in `check(...)` so a broken precondition fails loudly. Bonus: `lastDiceCount` is now correctly set in these fixtures, which the old bypass skipped.
- **`tryRerollThisTurn` behavior unchanged** — the diff only adds documentation about `lastDiceCount` being deliberately preserved; the SQL is identical.

**No correctness issues.** Two optional, non-blocking nits:

**1. Duplicated fixture incantation across test files.** The `check(gameDao.tryRecordDiceRoll(id, diceRoll = …, diceCount = …)) { "fixture: game not in ROLL_DICE" }` block (and the paired `tryTransitionPhase` for BUY_OR_BUILD) is now copy-pasted across `GameDaoTest`, `PurchaseServiceIntegrationTest`, `GameSyncServiceIntegrationTest`, and `SpringContextRestartRecoveryIntegrationTest`. A small shared test helper (e.g. `advanceToResolveEffects(gameId, roll, count)` / `advanceToBuyOrBuild(...)`) would remove the repetition and keep future fixture changes in one place. Fine to defer.

**2. PR description wording mismatch.** The body says the BUY_OR_BUILD setup uses `updateTurnPhase(id, BUY_OR_BUILD)`, but the code uses `tryTransitionPhase(RESOLVE_EFFECTS, BUY_OR_BUILD)`. The code is the better choice (it's guarded) — just update the description text.

Issue #433, opened 2026-06-30
TV_STATION steals from the bank instead of a chosen player


valentinaschiavon99:
## Summary

`TV_STATION` gives the active player 5 coins from the bank instead of stealing 5 coins from a chosen opponent.

## Location

`EarningsServiceImpl.kt:168–170`

## Root cause

`PaymentSource.CHOSEN_PLAYER` falls into the `bankEarned` branch, so the active player receives 5 coins from the bank rather than stealing 5 from a chosen opponent.

## Fix

Add a separate interaction round-trip to pick a target player, then deduct from that player's balance rather than from the bank.

---
_Reported in Machi Koro bug report, generated 2026-06-29._

PR #440, opened 2026-06-30
434 business center does nothing income 0 paymentsourcenone unhandled

### LINKED ISSUE #434: BUSINESS_CENTER does nothing (income 0, PaymentSource.NONE unhandled)
## Summary

`BUSINESS_CENTER` does nothing. It is seeded with `income = 0` and earns 0 coins at runtime. Players pay 8 coins for a card with no effect.

## Location

- `DataSeeder.kt:127–133`
- `PaymentSource.kt:8`

## Root cause

The card is seeded with `income = 0`. `PaymentSource.NONE` exists but is never handled in the resolve flow, so the card-swap mechanic is never triggered.

## Fix

Implement a dedicated action endpoint for the card-swap mechanic and handle `PaymentSource.NONE` in the resolve flow.

---

eles17:
## Summary

Fixes the `BUSINESS_CENTER` card behavior so it no longer acts as a no-op after activation.

Previously, `BUSINESS_CENTER` was seeded with `income = 0` and treated like a normal purple income card path. Because its effect is a card swap rather than a coin transfer, resolving effects after rolling a 6 did not produce any useful gameplay behavior.

This PR adds a dedicated Business Center swap action, corrects the card metadata, and covers the new behavior with focused tests.

## What Changed

- Added a new WebSocket action:
  - `/app/game.businessCenter.swap`

- Added `BusinessCenterSwapRequest` with:
  - `gameId`
  - `targetPlayerId`
  - `offeredCardType`
  - `requestedCardType`

- Implemented Business Center swap logic in `EarningsServiceImpl`.
  - Only usable by the active player.
  - Only usable after effects have been resolved for a dice roll of `6`.
  - Requires the active player to own `BUSINESS_CENTER`.
  - Requires the target player to be a different player in the same game.
  - Requires both players to own the cards being exchanged.
  - Rejects swaps involving major establishments.
  - Rejects swapping the same card type with itself.

- Updated purple-card earnings resolution so `PaymentSource.NONE` is not treated as bank income.

- Corrected `BUSINESS_CENTER` seed data:
  - Changed `paymentSource` from `CHOSEN_PLAYER` to `NONE`.

- Added a Flyway migration:
  - `V8__business_center_payment_source_none.sql`
  - Updates existing databases so `BUSINESS_CENTER` uses `payment_source = 'NONE'`.

- Updated WebSocket API documentation in `OpenApiConfig`.

## Tests Added / Updated

- Added service tests for:
  - `PaymentSource.NONE` purple cards not awarding coins.
  - Successful Business Center card swap.
  - Rejecting swaps involving major establishments.

- Added controller tests for:
  - Successful Business Center swap broadcast.
  - Business Center swap rejection broadcast.

- Updated seed-related tests and test seed data to expect `PaymentSource.NONE`.

## Verification

Ran successfully:

```bash
./gradlew test --tests org.machikoro.server.service.EarningsServiceImplTest --tests org.machikoro.server.controller.EarningsControllerTest
./gradlew compileKotlin compileTestKotlin

### DIFF
--- a/src/main/kotlin/org/machikoro/server/config/OpenApiConfig.kt
+++ b/src/main/kotlin/org/machikoro/server/config/OpenApiConfig.kt
@@ -45,6 +45,7 @@ class OpenApiConfig {
     | `/app/game.start` | `StartGameRequest` | `GAME_STARTED`, `GAME_ACTION` to `/topic/game/{gameId}` |
     | `/app/game.rollDice` | `RollDiceRequest` | `ROLL_DICE`, `GAME_ACTION` to `/topic/game/{gameId}` |
     | `/app/game.resolveEffects` | `ResolveEffectsRequest` | `GAME_ACTION` to `/topic/game/{gameId}` |
+    | `/app/game.businessCenter.swap` | `BusinessCenterSwapRequest` | `GAME_ACTION`, or `ERROR` with `payload.context.event = BUSINESS_CENTER_SWAP_FAILED`, to `/topic/game/{gameId}` |
     | `/app/game.advancePhase` | `AdvancePhaseRequest` | `DIRECT_PHASE_ADVANCE_FORBIDDEN` to `/user/queue/errors`; no state changes |
     | `/app/game.purchase` | `PurchaseRequest` | `GAME_ACTION`, or `ERROR` with `payload.context.event = PURCHASE_FAILED`, to `/topic/game/{gameId}` |
     | `/app/game.endTurn` | `EndTurnRequest` | `GAME_ACTION` or `GAME_END` to `/topic/game/{gameId}` |
--- a/src/main/kotlin/org/machikoro/server/controller/EarningsController.kt
+++ b/src/main/kotlin/org/machikoro/server/controller/EarningsController.kt
@@ -3,6 +3,7 @@ package org.machikoro.server.controller
 import io.github.springwolf.core.asyncapi.annotations.AsyncListener
 import io.github.springwolf.core.asyncapi.annotations.AsyncOperation
 import org.machikoro.server.auth.requireUserPrincipal
+import org.machikoro.server.dto.BusinessCenterSwapRequest
 import org.machikoro.server.dto.MessageType
 import org.machikoro.server.dto.ResolveEffectsRequest
 import org.machikoro.server.dto.WebSocketErrorDto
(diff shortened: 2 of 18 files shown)


valentinaschiavon99 (dismissed):
## Review

**What it does:** Fixes the `BUSINESS_CENTER` card, which was a no-op (seeded `income = 0`, `paymentSource = CHOSEN_PLAYER`). Adds a dedicated `/app/game.businessCenter.swap` WebSocket action + `BusinessCenterSwapRequest` DTO, implements `swapBusinessCenterCard` in `EarningsServiceImpl`, corrects the seed/`payment_source` to `NONE` (with Flyway `V8` migration), excludes `NONE` purple cards from bank income, and adds service + controller tests.

The implementation is solid overall — validation is thorough (active-player-only, valid target, ownership of both cards, no major establishments, no same-type swap), the four `upsert` calls are correct (distinct player/card pairs, and `upsert` deletes rows at quantity ≤ 0 so no ghost cards), and tests are focused. One real gameplay bug survives review.

### Findings

**1. `EarningsServiceImpl.swapBusinessCenterCard` — no once-per-activation guard; the swap can be performed unlimited times per turn.**
The only gate is `turnPhase == BUY_OR_BUILD && lastDiceRoll == 6`, and the method never records that the swap was used or advances the phase. `lastDiceRoll` stays `6` for the entire `BUY_OR_BUILD` window, so the active player can call `/app/game.businessCenter.swap` repeatedly and swap an arbitrary number of cards in a single turn. In Machi Koro a Business Center activation grants exactly **one** swap. A flag like the existing `hasPurchasedThisTurn`/`rerolledThisTurn` (e.g. a `businessCenterUsedThisTurn`) would close this. Worth a test covering a rejected second swap.

### Note (not blocking, coordination)
The purple-filter rewrite in `processPurpleCards` re-introduces `paymentSource == CHOSEN_PLAYER` into `bankEarned`. Against the current `main` this is behavior-preserving (the stated goal — excluding `NONE` — is met, and Business Center is income 0 anyway). But it touches the **exact line** the in-flight #433 branch rewrites in the opposite direction — that branch removes `CHOSEN_PLAYER` from `bankEarned` so TV Station steals instead of minting free bank coins. Expect a merge conflict, and make sure whichever lands second keeps `CHOSEN_PLAYER` **out** of `bankEarned`. Since the real fix here is the swap action (and Business Center is income 0), consider not whitelisting `CHOSEN_PLAYER` at all.

---

## Verdict: Request changes

Well-structured and well-tested, but the missing once-per-turn guard (#1) lets the active player swap an unlimited number of cards in a single Business Center activation, which is a real rule violation. Add a per-turn guard (and ideally a test covering a rejected second swap), and reconcile the purple-filter line with #433. Everything else looks good to merge.

valentinaschiavon99 (approved):
lgtm

PR #441, opened 2026-06-30
fix: TV Station steals from a chosen opponent (#433)

### LINKED ISSUE #433: TV_STATION steals from the bank instead of a chosen player

valentinaschiavon99:
## Summary

Closes #433.

`TV_STATION` was handing the active player **5 coins from the bank** instead of letting them **steal 5 coins from a chosen opponent**.

### Root cause
In `EarningsServiceImpl.processPurpleCards`, the bank-income branch filtered purple cards with `paymentSource != ALL_PLAYERS`. `PaymentSource.CHOSEN_PLAYER` (TV Station) therefore fell into `bankEarned`, so the active player simply received the card's income from the bank — no opponent was ever charged.

### Fix
A steal from a *chosen* opponent can't be resolved in the same pass that distributes dice earnings — the server doesn't yet know who the victim is. So the resolution is split into a dedicated interaction round-trip, modelled with a new turn phase:

```
RESOLVE_EFFECTS --resolveEffects--> AWAIT_TV_TARGET --chooseTvStationTarget--> BUY_OR_BUILD
```

- **`processPurpleCards`** now banks only `PaymentSource.BANK` purple income. CHOSEN_PLAYER no longer mints free coins. (No standard purple card is bank-sourced, so this branch is now effectively a guard; the steal is handled separately.)
- **`resolveEffects`** detects a pending TV Station steal (active player activated a TV Station **and** at least one opponent has coins) and parks the turn in the new **`AWAIT_TV_TARGET`** phase instead of advancing to `BUY_OR_BUILD`. If no opponent has coins, the steal would be a no-op and the round-trip is skipped entirely.
- **New endpoint `/app/game.chooseTvStationTarget`** (`TvStationTargetRequest { gameId, targetPlayerId }`) → **`resolveTvStationTarget`** applies the steal: the active player takes `5 × (TV Station count)` from the chosen opponent, **capped at that opponent's balance** so no coins are created, then advances to `BUY_OR_BUILD`.

### Concurrency / idempotency
The destination phase is decided *before* the optimistic `tryTransitionPhase` CAS, so a duplicate `resolveEffects` still loses the phase race and cannot double-apply earnings. `resolveTvStationTarget` likewise transitions out of `AWAIT_TV_TARGET` before moving any coins, so a duplicate target choice is rejected with `EFFECTS_ALREADY_RESOLVED`.

## Client contract

| Phase after `resolveEffects` | Meaning | Next client action |
|---|---|---|
| `BUY_OR_BUILD` | No TV Station pending (unchanged flow) | proceed to buy/build |
| `AWAIT_TV_TARGET` | Active player must pick a victim | send `/app/game.chooseTvStationTarget` with the opponent's `playerId` |

`chooseTvStationTarget` broadcasts `GAME_ACTION` with `payload.event = TV_STATION_RESOLVED` on success, or `ERROR` with `payload.context.event = TV_STATION_FAILED`. New rejection codes: `NO_PENDING_TV_STATION`, `INVALID_TV_STATION_TARGET`. Eligible targets are the opponents with `coins > 0` in the broadcast state snapshot.

## Database

`V8__add_await_tv_target_turn_phase.sql` updates the `games_turn_phase_check` constraint to allow `AWAIT_TV_TARGET` (fits the existing `VARCHAR(20)` column).

## Notes
- `BUSINESS_CENTER` is also seeded as `CHOSEN_PLAYER` but with income `0`, so it contributes nothing to the steal amount and never triggers the round-trip — its card-swap mechanic remains out of scope (tracked separately under #434).

## Tests
All 749 tests pass. Added coverage:
- **Unit (`EarningsServiceImplTest`)** — TV Station no longer pays from the bank; `resolveEffects` parks in `AWAIT_TV_TARGET` / skips when opponents are broke / rejects re-resolution; `resolveTvStationTarget` steals, caps at the victim's balance, handles a broke target, and rejects invalid targets, wrong phase, and stale transitions.
- **Controller (`EarningsControllerTest`)** — `chooseTvStationTarget` broadcasts success, domain rejections, and propagates `NOT_YOUR_TURN`.
- **Integration (`EarningsIntegrationTest`)** — full DB-backed round-trip: defer on `resolveEffects`, then steal (capped) on `resolveTvStationTarget`.


eles17 (dismissed):
Requesting changes.

There is one merge-blocking gameplay edge case:

When the active player owns both Stadium and TV Station, both activate on roll 6. `resolveEffects` decides whether to enter `AWAIT_TV_TARGET` before `processEarnings` applies Stadium. If Stadium drains all opponents to 0 coins, the game still parks in `AWAIT_TV_TARGET`, but the client contract says eligible TV Station targets are opponents with `coins > 0`.

That can leave the turn stuck waiting for a target the client should not offer.

Please adjust the flow so the TV Station pending check is based on post-non-TV-earnings balances, or make broke opponents valid no-op targets consistently in both backend validation and client contract.

Suggested test case:
Active player owns `STADIUM` + `TV_STATION`, rolls 6, opponents have only enough coins to be drained by Stadium. After `resolveEffects`, the game should advance to `BUY_OR_BUILD`, not `AWAIT_TV_TARGET`.

valentinaschiavon99 (commented):
## Requesting changes

Solid, well-tested fix for #433. The round-trip design is clean and the concurrency model is correct: deciding the destination phase *before* the optimistic `tryTransitionPhase` CAS means a duplicate `resolveEffects` still loses the phase race, and `resolveTvStationTarget` transitions out of `AWAIT_TV_TARGET` before moving any coins. I confirmed the new phase is properly locked — `endTurn` and `purchase` gate on `BUY_OR_BUILD` and `rollDice` on `ROLL_DICE`, so the active player can't bypass the steal while parked in `AWAIT_TV_TARGET`. Migration is correctly numbered `V10`.

Two minor, non-blocking observations (follow-up, not required for merge):

1. **Redundant DB queries on the hot path.** `isTvStationStealPending` runs on *every* `resolveEffects`, and its `tvStationStealAmount` re-issues `cardDao.findByActivationNumber(diceRoll)` and `playerCardDao.findByPlayerId(activePlayerId)` — both already computed in `calculateFinalCoins` immediately above. Could be derived from the existing `matchedCardsByPlayer` to save two round-trips per resolution.

2. **A 0-coin opponent is accepted as a target.** `resolveTvStationTarget` only validates "opponent and not the active player". If both a rich and a broke opponent exist, sending the broke `playerId` no-ops the steal and still advances to `BUY_OR_BUILD`. The PR contract says eligible targets have `coins > 0`, but the authoritative server doesn't enforce it. Defensible under the cap-at-balance design — just flagging it as a deliberate choice rather than an accident.

Nice test coverage on the deferral, the cap, the broke-target/Stadium-drain skips, and the stale-transition races.

valentinaschiavon99:
## Review

This is a clean, well-tested fix and the core mechanics hold up under scrutiny — concurrency (optimistic CAS before any coin movement), no permanent stall in `AWAIT_TV_TARGET`, and the Stadium-drain edge case are all handled correctly. Two cleanup items are worth addressing before merge.

### Findings

**1. `EarningsServiceImpl.kt:43` — `processEarnings` is now orphaned in production.**
Before this PR, `resolveEffects` called `processEarnings`; the rewrite has `resolveEffects` call `calculateEarnings` + `applyCoinChanges` directly instead. `processEarnings` (and its `EarningsService` interface entry) is now referenced only by tests. Either remove it, or — if it's intended public API — note why it's retained. Otherwise it's dead surface that future readers will assume is live.

**2. `EarningsServiceImpl.kt:93 & 407` — steal amount is computed two different ways.**
`calculateEarnings` derives `tvStationStealAmount` from `matchedCardsByPlayer`, while the standalone `tvStationStealAmount(diceRoll, activePlayerId)` helper recomputes it by re-querying `cardDao`/`playerCardDao`. The recomputation across the round-trip is necessary (separate transactions), but having two independent implementations of "sum CHOSEN_PLAYER income for the active player" is a divergence risk — if one is later changed (e.g. a Shopping-Mall-style modifier), the pending check and the applied steal could disagree. Consider extracting one shared computation over a passed-in card list.

### Note (non-blocking)
The PR description references the migration as `V8__add_await_tv_target_turn_phase.sql`, but the committed file is `V10__add_await_tv_target_turn_phase.sql`. The code is correct (V10 is the next free version; no collision with V8/V9 on the branch) — just stale text in the description.

One thing to be aware of, not a defect: with multiple TV Stations the implementation sums `5 × count` and takes it all from a single chosen opponent, whereas official rules let each TV Station target a different player. That's consistent with the scope of #433 and fine to leave for later.

valentinaschiavon99:
Reviewed for correctness, removed-behavior, and cross-file impact — LGTM (posting as a comment since GitHub blocks self-approval).

- **Enum insertion is safe:** `Tables.kt` persists `turn_phase` via `enumerationByName` (string, not ordinal), and all phase checks elsewhere (Purchase/GamePhase/Dice services, `tryRerollThisTurn`) are equality-based, so `AWAIT_TV_TARGET` correctly blocks purchase/endTurn/reroll while parked.
- **Idempotency holds:** destination phase is chosen before the CAS, all inside one transaction, so a duplicate `resolveEffects`/`chooseTvStationTarget` loses the race and rolls back without double-applying.
- **No liveness regression:** no turn-timeout/forfeit path assumes `BUY_OR_BUILD`.
- **Thorough test coverage** (cap-at-balance, broke/self/invalid target, wrong phase, stale transition, stadium-drain skip). CI green.

Minor (non-blocking): the PR description's Database section says `V8__…` but the actual migration is correctly numbered `V10` — worth a one-line description fix.

PR #445, opened 2026-06-30
Fix/436 Exceptions silently swallowed

### LINKED ISSUE #436: Uncaught exceptions silently swallowed for all other players (error not broadcast to game topic)
## Summary

Uncaught exceptions are silently swallowed for all other players. A non-`CustomWebSocketException` (e.g. a DB failure mid-roll) is delivered privately to the sender only, leaving every other player's UI in an unknown state.

## Location

`GameController.kt:372–379`

## Root cause

The fallback `catch (e: Exception)` block that would broadcast an `ERROR` to the game topic is commented out. Any non-`CustomWebSocketException` is delivered privately to the sender only — every other player's UI is left in an unknown state.

## Fix

Uncomment or replace with a proper broadcast of the error to the game topic.

---

Moritz2k1:
Uncommented catch block in the GameController


eles17 (approved):
LGTM. Approved.

